Issue #2, opened 2020-03-26
E-Mail overflow issue in pickachu theme


BerniWittmann:
Hey, 
first of all: you built a very cool and interactive Resume builder 👍 

I just found a small overflow issue in the Pickachu theme, where a long email overflows into the main content area. See the following picture:

<img width="250" alt="Bildschirmfoto 2020-03-26 um 17 27 43" src="https://user-images.githubusercontent.com/17594215/77671067-596ff900-6f87-11ea-8f4f-83134785ec8b.png">

Maybe the side area needs to be a little bit bigger and the overflow handling adjusted

AmruthPillai:
Thank you for raising this issue. Will get it fixed as soon as possible :)

BerniWittmann:
No worries, just wanted to let you know :)

AmruthPillai:
This has been fixed in #3, thanks to @viktorrudi's contribution.
And has also been deployed to production.

Thanks, everyone! :)

PR #3, opened 2020-03-26
Add word break for contact information in each template

viktorrudi:


AmruthPillai:
Woah, thank you so much! :) I was just about to do the same thing, you got the right Tailwind CSS class too. Thank you for fixing issue #2!

Issue #9, opened 2020-03-28
App crashes when importing existing resume.


Panzki:
Hey,
I checked the app on Thursday and started to create a resume. I exported my progress and wanted to continue today. When importing the exported state the app crashed. The reason is that some sections of the resume, I did not fill out (e.g. the language section), where missing in the export.
It seems that this this issue does not affect newer exports. 
<details>
<summary>Here is a sample export that breaks the app.</summary>
<p>

```json
{
    "data": {
        "profile": {
            "heading": "Profile",
            "photo": "https://i.imgur.com/Icr472Z.jpg",
            "firstName": "Nancy",
            "lastName": "Jackson",
            "subtitle": "Customer Sales Representative",
            "address": {
                "line1": "3879 Gateway Avenue",
                "line2": "Bakersfield,",
                "line3": "California, USA"
            },
            "phone": "+1 661-808-4188",
            "website": "nancyontheweb.com",
            "email": "nancyjack43@gmail.com"
        },
        "objective": {
            "enable": true,
            "heading": "Professional Objective",
            "body": "To obtain a job within my chosen field that will challenge me and allow me to use my education, skills and past experiences in a way that is mutually beneficial to both myself and my employer and allow for future growth and advancement."
        },
        "work": {
            "enable": true,
            "heading": "Work Experience",
            "items": [
                {
                    "id": "a208ec03-76e3-4428-ac5b-e17c3de4ac18",
                    "title": "On Point Electronics, NYC, NY",
                    "role": "Customer Service Representative",
                    "start": "Jan 2013",
                    "end": "July 2018",
                    "description": "- Organized customer information and account data for business planning and customer service purposes.\n- Created excel spreadsheets to track customer data and perform intense reconciliation process.\n- Received 97% positive customer survey results.\n- Speed on calls was 10% above team average.  \n**Key Achievement:** Designed and executed an automatized system for following up with customers, increasing customer retention by 22%."
                },
                {
                    "id": "bd8649f2-42d1-4424-acaf-a02c08c3322c",
                    "title": "Excelsior Communications, NYC, NY",
                    "role": "Customer Service Representative",
                    "start": "Oct 2009",
                    "end": "Dec 2012",
                    "description": "- Worked as a full time customer service rep in a high volume call center.\n- Received \"Associate of the Month\" award six times.\n- Chosen as an example for other associates in trainings.  \n**Key Achievement:** Received Customer Appreciation bonus in three of four years."
                },
                {
[...]

PR #10, opened 2020-03-28
Use initial state for section missing in an import.

Panzki:
This fixes  issue #9 by using the data from the initial state for sections missing in the imported data.


AmruthPillai:
Wow, I anticipated this error, but I was in a rut on wondering how to fix this elegantly. This solution is perfect, it would make sure all imports are up to date as well.

Merging it immediately. Thank you so much for quickly debugging and solving the issue! :)

PR #17, opened 2020-03-28
Netlify button

hwang381:
Hi there 👋 Great site and beautiful resumes that it can generate :)

To make it easier for people to deploy their own site I added a "Deploy to Netlify" button in this PR (I am not affiliated with the Netlify. I just like their platform a lot). Unfortunately since the button deploys the site from the master branch of **your** repo, that means I can only test whether the button works correctly (I think it should) after you merge my branch. I will make sure I test it out if you decide to do so :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,12 @@ I'm glad you asked :) Here is a step-by-step procedure on getting you started wi
 5. `npm start` - to start the development server
 5. `npm run build` - to build a production version of the app
 
+## How do I deploy the project myself
+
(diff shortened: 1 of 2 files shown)


AmruthPillai:
Hey there!

I'm a huge fan of Netlify too, and I think it's a great idea to allow users to host their own instance instantly if they have a Netlfy account. Except for a small grammar change here and there, I don't see any problem with the PR, the deployment should work as well, will test it out and make changes if necessary.

Thank you so much for your contribution to making this project better :)
Really really appreciate it!

PR #38, opened 2020-03-31
Quick dockerfile implementation

benjaminchodroff:
Awesome project. Saw the request for docker on reddit and thought this may help others. I see I am now 2 hours late... but perhaps my pull request will help you? 

You can install docker, and test by running these commands:
docker build -t reactive-resume .
docker run -p 5000:5000 reactive-resume:latest

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,2 @@
+node_modules
+npm-debug.log
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,21 @@
+# specify the node base image with your desired version node:<version>
+FROM node:10
+# replace this with your application's default port
+EXPOSE 5000
+# Create app directory
+WORKDIR /usr/src/app
+# Install app dependencies
+# A wildcard is used to ensure both package.json AND package-lock.json are copied
+# where available (npm@5+)
+COPY . ./
+
+RUN npm install
+RUN npm install -g serve
+RUN npm run-script build
+RUN npm run docs:build
+USER node
+# Production CMD
+CMD [ "serve", "-s", "build" ]
+# Comment out Production CMD and uncomment following for development
+#CMD [ "npm", "build" ]
+


AmruthPillai:
Oh, I am so sorry. I started working on it as soon as I created an issue on GitHub, the only thing slowing me down was my internet connection, but I have managed to successfully push a docker image and the changes required for the same.

You can read more about it in the documentation:
https://docs.rxresu.me/deployment/#docker

I'm really sorry that you had to go through all that trouble. I'll get better with making clear updates on what I'm working on further so that there won't be any clashes, but I do appreciate your efforts a lot and I do hope you keep contributing to the project :) Thank you so so much!

benjaminchodroff:
No problem at all! I'm glad you are pushing forward with this! It's awesome :)
Just some suggestions though - avoid multiple FROM (https://stackoverflow.com/questions/33322103/multiple-froms-what-it-means), declare a USER (don't run as root), and it would be best to leave nginx out of your dockerfile and instead create a docker-compose which uses two separate containers (it seems weird compared to how we used to think about virtual machines, but many want as little as possible in each container). But honestly - great work! Thank you!!!

AmruthPillai:
Yes, I just saw your comment on Reddit. I'm not quite sure on how to do this, as I'm still learning all about Docker. Could you reopen a PR and work around with the files I've made to make them better according to your experience? I would want to strive for maximum efficiency too, and dangling containers are a pain.

Thank you so much :) Please keep supporting, and I'll keep doing whatever I can to make the app more and more useful.

AmruthPillai:
Hey @benjaminchodroff, would you be making that extra fix or should we let this be? :)

benjaminchodroff:
Looks great to me. The nginx+node appears to work well - and is running as nginx. Let's go with what you got!

82jax:
Is the docker/portainer deployment code avail for this then as it's not obvious from an initial look?  Maybe I'm just missing something.

PR #51, opened 2020-04-01
[Workaround] Gengar: Break line when CertificationItem p content is too long.

pmop:
Hey there! Cool app, I'm already making use of it.

Regarding Gengar theme. Content in CertificationItem wont wrap if too long, which results in ugly looks, if you put something too long there, like an URL for instance.
As a work around, I've put tailwind's "break-words" there, which seems to fix for CertificationItem. 

Original:
Original
Fixed:
Fixed


AmruthPillai:
Thank you :)

PR #57, opened 2020-04-01
add mouse wheel scroll into tabBar

qiandongyq:
add mouse wheel scroll into TabBar component for a better user experience

### DIFF
--- a/src/shared/TabBar.js
+++ b/src/shared/TabBar.js
@@ -16,6 +16,8 @@ const TabBar = ({ tabs, currentTab, setCurrentTab }) => {
     }
   };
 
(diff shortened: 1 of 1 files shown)


AmruthPillai:
Thanks for the implementation, looks good and a much-needed UX feature at that :) Since I use the MacBook trackpad, I haven't tried it out using a mouse.

AmruthPillai:
I might have to revert this change for a while. While the fix is great for mice that have a scroll wheel, the onWheel function is causing a sort of lag/delay when swiping left/right on a laptop trackpad (both Mac and Windows. I'll look into some other methods here where scrolling on the wheel will scroll the tab, without causing side effects for other users.

PR #63, opened 2020-04-02
fix typo in Contributing

hatemhosny:
This PR fixes a typo in Contributing

great work by the way 👍

---
This change is [Reviewable](https://reviewable.io/reviews/amruthpillai/reactive-resume/63)

### DIFF
--- a/docs/contributing/README.md
+++ b/docs/contributing/README.md
@@ -30,7 +30,7 @@ Something that's missing on the app that's halting your progress from making the
 
 ## Translation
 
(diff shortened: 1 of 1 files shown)


AmruthPillai:
Oh, I am just embarrassed 🤦‍♂ 

I tend to type quickly and not read through what I've written, and this is the result of that. Thank you so much for finding this typo and fixing it, very much appreciated. I care about these things a lot too btw. Please do let me know if there any issues at all, would be open to hearing them :)

Thank you for this and your continued support!

PR #81, opened 2020-04-06
update dockerignore to exclude other unneeded folders/files

lpmi-13:
add additional files/folders that dont need to be in the Docker image


AmruthPillai:
Good catch :)

PR #126, opened 2020-04-21
Add Vietnamese translation

PhuNH:
The translation is based on currently in use l10n strings, not the old ones left in those json files.


AmruthPillai:
Thank you so much for your contribution :)

PR #127, opened 2020-04-21
Add level.label string

PhuNH:
Currently the label is displayed as "languages.level.label".

### DIFF
--- a/src/i18n/locales/en/leftSidebar/languages.json
+++ b/src/i18n/locales/en/leftSidebar/languages.json
@@ -2,6 +2,9 @@
   "key": {
     "label": "Name"
   },
+  "level": {
+    "label": "Level"
+  },
   "rating": {
     "label": "Rating"
   }
--- a/src/i18n/source/leftSidebar/languages.json
+++ b/src/i18n/source/leftSidebar/languages.json
@@ -2,6 +2,9 @@
   "key": {
     "label": "Name"
   },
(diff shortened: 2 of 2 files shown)


AmruthPillai:
Actually, this is the wrong file to be editing. You'll have to edit src/i18n/source/* files instead. If you can modify the PR to edit that file instead, I'll merge this :) Thank you for pointing that out, missed out this.

PhuNH:
I'm not really sure what that `src/i18n/source/` folder is for, seeing nothing mentions it in the code, I even removed it and got the app working normally (so far).
In `src/i18n/index.js` you have the list of languages with the fallback language being `en`.
And I also tested before committing. Now I've just tried changing `src/i18n/source/leftSidebar/languages.json` instead of the file changed in this commit `src/i18n/locales/en/leftSidebar/languages.json`, the label is still "languages.level.label".

AmruthPillai:
Ah, let me explain. The `i18n/source` folder isn't for the application, but for the translation management platform, Crowdin. It monitors the new strings in the application so that it can transfer the same to all other languages.

PhuNH:
So should files in both folders be updated?

AmruthPillai:
Yes, you can edit both files. Eventually, the Crowdin platform will be the single source of truth.

AmruthPillai:
All good, can be merged now. Thank you :)

PR #128, opened 2020-04-21
Add Vietnamese to the list as "Tiếng Việt"

PhuNH:
I suggest using the language names as written in their own languages, e.g. German -> Deutsch, Korean -> 한국어, Vietnamese -> Tiếng Việt, etc.


AmruthPillai:
That's a good point, to use the official name instead of its English name. Will make the change for other languages as well, thank you so much.

PR #170, opened 2020-05-19
Added Option to add GitHub and LinkedIn profile in the Profile Section.

ShubhamVerma1811:
Added Option to add GitHub and LinkedIn profile in the Profile Section.
But, as Material Icons don't have GH and LinkedIn icons I've added temporary icons from the website

### DIFF
--- a/src/assets/demo/data.json
+++ b/src/assets/demo/data.json
@@ -13,6 +13,8 @@
       },
       "phone": "+1 661-808-4188",
       "website": "nancyontheweb.com",
+      "github":"JaneDoe43",
+      "linkedin":"JaneDoe",
       "email": "nancyjack43@gmail.com"
     },
     "objective": {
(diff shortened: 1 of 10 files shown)


AmruthPillai:
First of all, thank you so much :) This is really great work.

I just pulled in the latest code and I was faced with this error:
```
Error: ContactItem(...): Nothing was returned from render. This usually means a return statement is missing. Or, to render nothing, return null.
```

I'm guessing this is because the JSON in my local storage doesn't have the github or linkedin field. This also needs to be handled before I can merge this so that it doesn't break functionality for those existing users :) I'm also looking into the code on how to do this the best way with minimal changes to code. Let me know if you have any suggestions?

ShubhamVerma1811:
Yep that missing fields are causing that.
I don't how the import and export code works but maybe you could check for
````
'github' && 'linkedin' in imported json
````

PR #177, opened 2020-06-06
fix: Reducer called twice due to use set on shallow copy of the state

SHENG-X:
- Using set on the shallow copy of the state object caused reducer called twice; reducer function should be pure; use deep copy of the state instead of shallow copy


SHENG-X:
fix reducer bug

PR #179, opened 2020-06-06
fix: Move and remove should match item by id

SHENG-X:
- Button does not respond to move and remove due to compare object instead of the id of the object

### DIFF
--- a/src/context/AppContext.js
+++ b/src/context/AppContext.js
@@ -92,31 +92,32 @@ const initialState = {
 
 const reducer = (state, { type, payload }) => {
   let items;
+  const stateCopy = JSON.parse(JSON.stringify(state));
 
   switch (type) {
     case 'migrate_section':
-      return set({ ...state }, `data.${payload.key}`, payload.value);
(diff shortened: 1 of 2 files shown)


src/utils/index.js
```diff
@@ -3,7 +3,7 @@ import html2canvas from 'html2canvas';
 import * as jsPDF from 'jspdf';
 
 const move = (array, element, delta) => {
-  const index = array.indexOf(element);
```
SHENG-X, line 6: indexOf is good for primitive types; in this case, we are dealing with a complex object so accessing by id is more reasonable

src/context/AppContext.js
```diff
@@ -103,7 +103,7 @@ const reducer = (state, { type, payload }) => {
       return set({ ...newState }, `data.${payload.key}.items`, items);
     case 'delete_item':
       items = get({ ...newState }, `data.${payload.key}.items`, []);
-      remove(items, x => x === payload.value);
+      remove(items, x => x.id === payload.value.id);
```
SHENG-X, line 106: match items by id instead of matching them by the whole object

PR #180, opened 2020-06-06
Fix: move and remove items 

SHENG-X:
- Convert skill and hobby to objects, each object contains a unique key
- Fixed move and remove item does not work properly for eduction and skills page

### DIFF
--- a/src/components/LeftSidebar/tabs/Skills.js
+++ b/src/components/LeftSidebar/tabs/Skills.js
@@ -1,4 +1,5 @@
 import React, { useState, useContext } from 'react';
+import { v4 as uuidv4 } from 'uuid';
 
 import AppContext from '../../../context/AppContext';
 import Checkbox from '../../../shared/Checkbox';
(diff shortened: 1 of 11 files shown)


src/components/LeftSidebar/tabs/Skills.js
```diff
@@ -31,7 +32,7 @@ const SkillsTab = ({ data, onChange }) => {
       <hr className="my-6" />
 
       {data.skills.items.map((x, index) => (
-        <Item item={x} key={index} index={index} onChange={onChange} dispatch={dispatch} />
+        <Item item={x} key={x.id} index={index} onChange={onChange} dispatch={dispatch} />
```
SHENG-X, line 35: index prop probably not necessary since now we can use id to find the index. I'll key the index here for now. I'll let you make the call

src/utils/index.js
```diff
@@ -3,7 +3,7 @@ import html2canvas from 'html2canvas';
 import * as jsPDF from 'jspdf';
 
 const move = (array, element, delta) => {
-  const index = array.indexOf(element);
```
SHENG-X, line 6: indexOf is good for primitive types. we are dealing with objects so I changed it to use findIndex and use id as the key to match items

PR #197, opened 2020-06-22
Add A4 page break indicator

arsenikstiger:


AmruthPillai:
@arsenikstiger I like this addition, but it would've been much better if you had created a PR just for this one change. Instead, you've added the changes from the other PR as well on this, so I can't merge it directly. Could you raise a PR that has only this change?

arsenikstiger:
@AmruthPillai 
I fixed that problem with #199

PR #203, opened 2020-07-02
Updated TR translation

halilcakar:
I updated `i18n/locales/tr` files. 

Found out that on the `RightSideBar/tabs/About.js` it didn't work it should. 

I don't know much about React so i can not fix those but just letting you know maybe you or someone can fix it ☺

### DIFF
--- a/src/i18n/index.js
+++ b/src/i18n/index.js
@@ -71,6 +71,10 @@ const languages = [
     name: 'Tamil (தமிழ்)',
   },
   {
+    code: 'tr',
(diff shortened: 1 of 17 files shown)


halilcakar:
Oh also I saw that inside code `placeholder`'s hardcoded. Might be also nice to move them into `json` files maybe?

halilcakar:
So one more thing. Language does not go into localStorage. When i close and open it up, it's coming English. Maybe nice to fix that too :))

PR #230, opened 2020-07-19
fr translation

MeisterLLD:
I wrote fr.json. I left two items in english : sourceCode and translate


AmruthPillai:
@MeisterLLD Thank you so much, this is brilliant. I can fill the other two with a few of my friends who know French. Thank you so much for the help, will integrate this into the app immediately. The update should be available in few minutes of merging.

MeisterLLD:
You're welcome ! Thanks for this very nice app :) I've left those intentionally, as I guess, they make more sense in English, but it you think it is worth it I can translate those too.

PR #251, opened 2020-07-22
Customize skills level input

SophieMdl:
Replace skills level select by a simple input => allow more flexibility and all langage support for this fields


AmruthPillai:
LGTM :) Will get it merged into production soon.

PR #271, opened 2020-07-30
adding support for arabic language

A7madXatab:
adding support for arabic language, making it usable in the whole middle east


AmruthPillai:
Thank you so much for contributing to Reactive Resume! :)

PR #275, opened 2020-07-31
Add Norwegian translation

rubjo:
...and a single English typo I came across.

BTW, should line 225 from the translation say "LinkedIn" instead of "Reactive Resume"?


AmruthPillai:
Haha, thank you for going the extra mile and adding it to the i18n locales as well :) This is brilliant, thank you so so much for your contribution.

Issue #331, opened 2020-09-13
Issue while adding new  experience


Abhilash9423:
while clicking on the add experience button the web page is blanking out

adityag81:
Same here.

> 
> 
> while clicking on the add experience button the web page is blanking out

Radikad:
same for me too.

Radikad:
hi @Abhilash9423 , @adityag81 

I've found the error and a way to avid it. 

in the below files there is a prop called **hasDate**

/src/components/builder/left/sections/Education.js
/src/components/builder/left/sections/Work.js

by commenting that you can avoid the errors. (this is not a fix but a way to avoid) 

```
       <List
        //hasDate
        path={path}
        event={event}
        titlePath="institution"
        textPath="field"
      />
```

hope this help.

robert-w-gries:
@Radikad your workaround works because there's an issue with how the dates are formatted. See #301

AmruthPillai:
This should be working now. Thanks to @robert-w-gries for the fix :)

Issue #301, opened 2020-08-13
Setting date in List causes 'Rendered more hooks than during the previous render' error


robert-w-gries:
I've been running into the following error while running the development server:

```
Uncaught (in promise) Error: Rendered more hooks than during the previous render.
    React 5
    useTranslation useTranslation.js:52
    formatDateRange index.js:29
    List List.js:44
    List List.js:36
    React 18
    _callee$ DataModal.js:70
    tryCatch runtime.js:45
    invoke runtime.js:274
    method runtime.js:97
    Babel 2
react-dom.development.js:14980
```

The problem is that hooks are [meant to only run in the top level of React function components](https://reactjs.org/warnings/invalid-hook-call-warning.html#breaking-the-rules-of-hooks). As a workaround, I commented out the `useTranslation()` hook, which fixes the issue:

```diff
diff --git a/src/utils/index.js b/src/utils/index.js
index 156daef..8cfcd0c 100644
--- a/src/utils/index.js
+++ b/src/utils/index.js
@@ -27,14 +27,14 @@ export const formatDate = ({ date, language = 'en' }) => {
 };

 export const formatDateRange = ({ startDate, endDate, language = 'en' }) => {
-  const { t } = useTranslation();
+  // const { t } = useTranslation();
   const start = `${dayjs(startDate)
     .locale(language.substr(0, 2))
     .format('MMMM YYYY')}`;

   const end = dayjs(endDate).isValid()
     ? `${dayjs(endDate).locale(language.substr(0, 2)).format('MMMM YYYY')}`
-    : t('shared.forms.present');
+    : 'Present';

   return `${start} - ${end}`;
 };
```

AmruthPillai:
Could you suggest many any other alternatives to using the Translation here? Maybe passing it to the function from the calling component might help?

robert-w-gries:
>Maybe passing it to the function from the calling component might help?

Yeah I think this is the way to go. You already call the `useTranslation()` hook [here](https://github.com/AmruthPillai/Reactive-Resume/blob/develop/src/components/builder/lists/List.js#L24), just get a reference to the `t` function in that line and pass it down to `formatDateRange()`

AmruthPillai:
This should be fixed now, thanks to @robert-w-gries for the PR! :)

PR #341, opened 2020-10-02
Fix app crashes after setting date range

### LINKED ISSUE #336: Adding work experience causes the page to go blank
When entering some data into the work experience field, clicking the "add experience" button causes the entire webpage to go blank. I have to refresh to get back into the app. Sometimes, the data I submitted is saved, sometimes it isn't.

Encountered on Chrome 85.0.4183.121

robert-w-gries:
Fixes #336, #331, #301 

Hooks should only be called from the [top-level of React Function Components or custom Hooks](https://reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level), which is why there have been various crash issues reported.

This PR is a quick-fix to prevent severe crashes, but there is likely a more elegant solution that can be submitted in a follow-up PR

### DIFF
--- a/src/components/builder/lists/List.js
+++ b/src/components/builder/lists/List.js
@@ -21,7 +21,7 @@ const List = ({
   hasDate,
   event,
 }) => {
-  const { i18n } = useTranslation();
+  const { t, i18n } = useTranslation();
   const items = useSelector(path, []);
   const { emitter } = useContext(ModalContext);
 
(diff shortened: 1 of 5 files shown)


AmruthPillai:
Thank you for your contribution! :)

robert-w-gries:
You're welcome, thank you for creating this tool! I have gotten many compliments from interviewers regarding the layout + style of my resume

PR #374, opened 2020-11-08
Add a Birthdate field to the profile information

### LINKED ISSUE #314: [Feature Request]: Add a Birthdate field to the profile information
It could be optional to show it or not, but adding some kind of birthdate input could be good.

gianantoniopini:
Added new "birthDate" field to the Profile builder and the Castform template.
This new "Date of Birth" field will only be shown by the template if the user entered it.

Resolves #314


AmruthPillai:
Looks like a good PR, but just a few things I'd like to see. Happy that you already got the translations to other languages down, that helps a lot.

- Would you be able to export BirthDate into a separate component BirthDateA?
- Would it be possible to update all templates to incorporate the birth date field, not just Castform?
- Instead of passing down language to the component, you should be able to get language from `data.metadata.language`. This is done because language should be visible even when other non-logged in users visit the resume public link and should be able to see the same language.

gianantoniopini:
Thanks for creating this useful tool and for checking the PR 👍 
I will look into applying the changes you'd like to see. They should all be possible.

gianantoniopini:
@AmruthPillai Hi Amruth, I've now applied the changes you suggested. I hope that the birthday cake icon looks fine in the templates and not too "unprofessional"?  While modifying the Celebi template, I changed it from using the ContactE block to using the ContactC block. As a result of this change, the ContactE block is not used any longer and has been deleted. Hope this is fine for you.

AmruthPillai:
I always appreciate people who keep their code clean 🧹 
Thank you for the contribution, will merge it as soon as I'm done testing.

marinang:
Hi, is this already deployed?

gianantoniopini:
@marinang Hi, yes I believe this has been deployed with version 2.3

Issue #325, opened 2020-09-05
App is great but can't change font size


WilsonNet:
Great app, thanks for sharing, but I really think it would benefit from an option to edit the font size.

vijay-yadav-3:
When can we expect this feature. ??
PS. Great App

rpolley:
Going to take a stab at this.

AmruthPillai:
Closing this in favor of Reactive Resume v3 release.
If you continue to face this issue, please raise a new one and I'll have a look at it as soon as I can.

In the meantime, if you are able to, please consider donating to keep supporting the project: https://www.buymeacoffee.com/AmruthPillai :)

Thank you!

Issue #256, opened 2020-07-23
Unable to change font size in web app.


darshittrevadia:
I cannot seem to find an option to change the font size.  
How do I change the font size on the web app?

SophieMdl:
@AmruthPillai I could handle this one :)

AmruthPillai:
@SophieMdl Please do give it a shot :) I would very much appreciate that. I'm finding very less time to work on this as my work tasks are very taxing. If you could work on the same and raise a PR, it would be just great!

A7madXatab:
@SophieMdl  are you still working on this or i should do it ?

thanks.

SophieMdl:
Hi, sorry but I don’t have time currently. Thanks to handle this :)

AmruthPillai:
Closing this in favor of Reactive Resume v3 release.
If you continue to face this issue, please raise a new one and I'll have a look at it as soon as I can.

In the meantime, if you are able to, please consider donating to keep supporting the project: https://www.buymeacoffee.com/AmruthPillai :)

Thank you!

PR #413, opened 2020-12-26
Add ability to customize font size

### LINKED ISSUE #379: Font Size
Hi, is it possible to customize font-size? I couldn't find it. The default font is looking so small.

rpolley:
Added an ability to customize font sizes. This is set at a new section in the settings.
The controler for this sets a number of css variables which are used in css rules that
override tailwind's text-<size> to use the font size and line spacing contained in the
variable. The control itself is a simple logarithmic slider that controls a "scale
factor" by which all the text sizes in a resume are scaled by.

Fixes #379 #325 #256

### DIFF
--- a/src/components/builder/right/RightSidebar.js
+++ b/src/components/builder/right/RightSidebar.js
@@ -10,6 +10,7 @@ import Fonts from './sections/Fonts';
 import Layout from './sections/Layout';
 import Settings from './sections/Settings';
 import Templates from './sections/Templates';
+import FontSizes from './sections/FontSizes';
 
 const getComponent = (id) => {
   switch (id) {
@@ -27,6 +28,8 @@ const getComponent = (id) => {
       return Settings;
     case 'about':
       return About;
+    case 'font-sizes':
+      return FontSizes;
     default:
       throw new Error();
   }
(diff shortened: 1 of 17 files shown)


rpolley:
One thing this is currently missing is localizations for the setting name is called outside of the English version.

vijay-yadav-3:
@rpolley thanks for the PR. I was waiting for this for a long time. 
@AmruthPillai Can you merge this PR. This is a feature we all really need. BTW will checkout and try to run the application from this PR.

AmruthPillai:
Reviewing it now :)

vijay-yadav-3:
@rpolley @AmruthPillai I can see the font size slider and it is changing size on UI. But while exporting the resume the font size is still the same. Not changing in the final exported PDF. Am I the only one facing this issue?

rpolley:
@vijay-yadav-3 @AmruthPillai from what it looks like none of the settings currently get serialized at all when exporting the resume.

AmruthPillai:
I actually added the serializing into JSON portion of the code in a subsequent commit after @rpolley's PR. That should be a part of it, and from what I can wee, it does preserve font size when refreshing the page too. Is the font size even when they are on the `/r/:resumeId` page? That is the page used to export PDFs too, so it should be visible there.

AmruthPillai:
Fixed in v2.5.2 :)

PR #423, opened 2021-01-12
Unit testing environment set up and addition of some unit tests

gianantoniopini:
Set up the unit testing environment and added some initial unit tests.

The unit testing environment has been set up following the [Gatsby Unit Testing](https://www.gatsbyjs.com/docs/how-to/testing/unit-testing/) and [Gatsby Testing React Components](https://www.gatsbyjs.com/docs/how-to/testing/testing-react-components/) guides.
The "gatsby-plugin-firebase" dependency has been replaced by a stub implementation called "FirebaseStub" and placed under the new "\_\_mocks\_\_" directory.
Jest has been configured to collect coverage information and store it under the "test-coverage" directory.
The "test" script in package.json has been updated to run "jest" and is also run as prebuild step.

Some basic unit tests for the Dashboard and Builder components, and for the Castform template, have been added.

I would be keen to implement additional unit tests later on to try and cover more user interactions.


AmruthPillai:
This is a great addition, and one that is very much necessary now that the project is growing. Thank you for taking the time to do this. Will review this in the coming weekend if time permits and see if I can add some test cases as well. Thank you very much for this contribution! 💯

gianantoniopini:
Glad to know that this could be a useful addition. I also think that it is useful to have some unit tests which can be run automatically. Thanks! 👍

gianantoniopini:
Added basic implementation of data updates to FirebaseStub.database() . This should suffice to support most current usage of the Firebase.database() dependency in the app.
The Firebase Stub can anyhow be extended as more unit tests are implemented in the future, if these tests leverage Firebase functionality that has not yet been implemented in the stub. For example: currently only auth().signInAnonymously() has been implemented in the stub. If in the future we would need to also test the loginWithGoogle user interaction then we should implement auth().signInWithPopup() in the stub, or just mock the function in the unit test.

PR #448, opened 2021-02-08
Added Lithuanian translations.

agirdzius:
Found this tool yesterday and I think it's awesome. Decided to contribute by translating it to my native tongue.


AmruthPillai:
Wow, I appreciate the effort to make the extra changes to code to include the Language in the list :) Thank you so much for your contribution. Could I just bother you for one small change, to be able to increment the version by 0.0.1?

agirdzius:
Absolutely. Wasn't completely sure which version you are referring to, the `package.json` or `gatsby-config.js` one, but I've updated it in the latter. Ping me if it should be otherwise and I'll update.

PR #459, opened 2021-02-14
Dockerfile improvements

mlow:
- Add missing build dependencies for `sharp` node module (dependency of
  `gatsby-plugin-sharp`).
- Add ARG instructions for firebase variables, allowing them to be set
  via --build-arg flags when building the image.
- Use `npm ci` instead of `npm install`, and execute it after copying
  only package*.json. This prevents having to fetch all dependencies if
  package*.json hasn't changed between builds.
- Use `node:alpine` instead of `node` - not an improvement per-se,
  but I happen to be more familiar with adding build dependecies in the
  alpine based node image.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,29 @@
-FROM node as builder
-
+FROM node:alpine as builder
 WORKDIR /app
-COPY ./ /app/
 
-RUN npm install
+RUN apk update \
+    && apk add --no-cache \
+        g++ \
+        yasm \
+        make \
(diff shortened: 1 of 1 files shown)


AmruthPillai:
Thank you so much for this very insightful PR. I don't have much experience with docker, but I'm glad the project has you to help :) Hope this clears out some of the issues people have been facing with Docker and Firebase. We need a more straightforward guide/video on how to set this up...

Issue #323, opened 2020-09-03
[FEATURE REQUEST] Show/hide box on items such as work experience


ilangleben19:
I use different sets of work experience, awards, skills, etc. depending on where I'm applying. I'd love to be able to add all my experience/everything to one CV file, and to have a checkbox to show/hide each item. This would allow me to easily generate a specific set of items on a CV without having to duplicate one big template and manually remove items.

Thanks, great software! :)

AmruthPillai:
Closing this in favor of Reactive Resume v3 release.
If you continue to face this issue, please raise a new one and I'll have a look at it as soon as I can.

In the meantime, if you are able to, please consider donating to keep supporting the project: https://www.buymeacoffee.com/AmruthPillai :)

Thank you!

PR #462, opened 2021-02-25
feat: adding a toggle switch for left panel items

anu1097:
New UI -
Screenshot 2021-02-26 at 3 17 54 AM

Description - Saw a good first issue - https://github.com/AmruthPillai/Reactive-Resume/issues/323

Approach taken - Read through the code to understand the data flow. Not the cleanest solution, but does the job.
Thought about creating a generic function or modify safety-check function to avoid code duplication. 
But that would require significant changes hence went against it and used this approach.


anu1097:
@AmruthPillai can you have a look. I think its ready to be merged.

src/contexts/ResumeContext.js
```diff
@@ -68,6 +68,18 @@ const ResumeProvider = ({ children }) => {
           debouncedUpdateResume(newState);
           return newState;
 
+        case 'on_toggle_use_item':
+          items = get(state, payload.path);
+          index = findIndex(items, ['id', payload.value.id]);
+          if ('isVisible' in items[index]) {
+            items[index].isVisible = !items[index].isVisible;
+          } else {
+            items[index].isVisible = false;
```
amruthpillai, line 77: If `isVisible` doesn't exist in the data array, we should have all sections visible by default. Here, when a user tries to import/load an older version of the resume, they wouldn't see any of the sections.

AmruthPillai (changes requested):
This needs to be addressed before we can merge :)

src/contexts/ResumeContext.js
anu1097:
```
export const isItemVisible = (section) =>
  section && section.isVisible !== false;
```

I am using this logic to determine whether to display or not. if IsVisible is not present in existing section. This function will return true. So this should handle it.

PR #477, opened 2021-04-03
Fixed bug where multiple resumes are created if "Create Resume" button is spammed

### LINKED ISSUE #476: Spamming `Create Resume` button creates multiple new resumes. Fix also mentioned.
# Demo of the bug
rxresume-bug

# Where this bug exists
It was a little confusing for me to figure out how the `Create Resume` modal was working but AFAIK `Create` and `Edit` resume is a part of `src/modals/DataModal.js`. 

https://github.com/AmruthPillai/Reactive-Resume/blob/8628ca5475e4ae79bc28fe70d61f8cca32b4f0dd/src/modals/DataModal.js#L94

I am pretty sure, the fix is very simple - 
original: `onClick={() => onSubmit(values)}`
solution: `onClick={() => !loading && onSubmit(values)}`

We just don't call `onSubmit(values)` if `loading` is true. @AmruthPillai I can create the PR with the fix if you also think this is the solution for the bug and assign this issue to me.

ceoshikhar:
Fixes #476 

# Demo of the fix
rxresume-bug-fix-demo

### DIFF
--- a/src/modals/DataModal.js
+++ b/src/modals/DataModal.js
@@ -91,7 +91,7 @@ const DataModal = ({
     : title.create;
 
   const submitAction = (
-    <Button type="submit" onClick={() => onSubmit(values)}>
+    <Button type="submit" isLoading={loading} onClick={() => onSubmit(values)}>
       {loading ? t('shared.buttons.loading') : getTitle}
     </Button>
   );


AmruthPillai:
Awesome, LGTM! :) Thank you for the contribution

PR #485, opened 2021-04-27
Delete Account: fix for Firebase cloud function and error handling improvements

### LINKED ISSUE #466: Resume still viewable & public even if the account is deleted
Hi,

I believed the same concern was raised in the past https://github.com/AmruthPillai/Reactive-Resume/issues/253

This is still not fixed, I've tested only with anonymous login. The steps to reproduce are same.

Proof of concept:

The account with ID fqwtkm has been deleted but the resume is still viewable and public

https://rxresu.me/r/fqwtkm

Thank you!

gianantoniopini:
Fixes #466 
<br>
Changes applied to the `deleteUser` Firebase cloud function:
- Performance improvement: filtering by userId when retrieving the resumes from the database
- Added removal of the user node
- Deletions of resumes and user done through a single atomic update (all or nothing)

Changes applied to the UserContext `deleteAccount` function:
- Performing user re-authentication, if user is signed in with Google, as a very first step. This is done to prevent possible 'auth/requires-recent-login' errors later on, when `currentUser.delete()` is called. Should an error occur in the re-authentication, then the app will notify the user and will not proceed further
- Using `await` when calling the `deleteUser` Firebase cloud function. Should an error occur in the cloud function then the app will notify the user and will not proceed further

Changes applied to the Settings `handleDeleteAccount` function:
- Call to the UserContext `deleteAccount` function is done using `await` and with `try...catch` block in order to notify the user of any error

### DIFF
--- a/functions/index.js
+++ b/functions/index.js
@@ -10,41 +10,49 @@ function timeout(ms) {
   return new Promise((resolve) => setTimeout(resolve, ms));
 }
 
-async function asyncForEach(array, callback) {
-  for (let index = 0; index < array.length; index++) {
-    // eslint-disable-next-line no-await-in-loop
-    await callback(array[index], index, array);
+const deleteUserFunctionHandler = async (_, { auth }) => {
+  if (!auth) {
+    throw new functions.https.HttpsError(
+      'failed-precondition',
+      'The function must be called while authenticated.',
+    );
   }
-}
+
+  try {
+    // 1. Delete user resumes
+    const userId = auth.uid;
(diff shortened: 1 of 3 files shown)


gianantoniopini:
@AmruthPillai Could you have a look at this PR ?

AmruthPillai (approved):
LGTM, thank you so much for this awesome contribution! :)

gianantoniopini:
@AmruthPillai It's a pleasure to have the chance to contribute :-) Thank you for reviewing this.

PR #526, opened 2021-07-06
fix typo in placeholder dates

lpmi-13:
We could probably even update this to something a bit more recent like 2018, depending on the context, but happy to leave the decision up to you.


AmruthPillai:
Woah, didn't notice this. Good eye you got there. 2018 would be a better option. Could you perhaps change it to that so I can merge this PR as is?

lpmi-13:
updated to 2018.

PR #527, opened 2021-07-06
add information on setting up firebase to the readme

lpmi-13:
I had to think a bit about what values to put in the .env file, and once I figured it out, I thought it would be helpful to add that information explicitly in the readme so other developers could speed up local development.


lpmi-13:
and actually, it looks like `FIREBASE_MEASUREMENTID` isn't even necessary to run locally, so I can add a commit to remove that from the `.env.example` if you want.

PR #528, opened 2021-07-06
update Project data model to use "Present" if no end date selected

lpmi-13:
Sometimes, projects are ongoing, and it would be helpful to be able to show that, similar to how it's done in the Work section.

This is what's expected in the `formatDateRange` function, and it's more consistent with the way things are done in other parts of the application.


lpmi-13:
a demo of the updates working locally, both for new projects and editing old projects

working present as end date

PR #543, opened 2021-07-28
German locale corrections and improvements

kescherCode:


AmruthPillai:
Thank you for the corrections.

PR #547, opened 2021-08-08
1] sidebar toggle feature 2] ask for confirmation before delete operation

preetamm:
1] people faced trouble seeing their resume changes, so I added a feature to toggle both left and right sidebars so users can have a bigger view of their resume.

2] sometimes users might accidently delete any of their list items so asking for confirmation before deletion is a must so I have added that.


AmruthPillai:
Thank you very much for your contribution to Reactive Resume :)

PR #599, opened 2022-01-09
Allow links for certifications

### LINKED ISSUE #575: Embeddin links in the certification
Thank you for your great work. It is really beneficial in preparing resumes. However, it will be better If we become able to embed certificate links into our certification part. Thanks for a great platform again.

adamantike:
Add option to set a URL for each certification. This is useful, for example, when sharing verified certificates as proofs provided by issuers.

Fixes #575

### DIFF
--- a/src/modals/sections/AwardModal.js
+++ b/src/modals/sections/AwardModal.js
@@ -11,6 +11,7 @@ const initialValues = {
   title: '',
   awarder: '',
   date: '',
+  url: '',
   summary: '',
 };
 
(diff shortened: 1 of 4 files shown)


adamantike:
I've rollbacked the changes to Awards, in case this project is looking forward to be compatible with https://jsonresume.org/, where only `certificates` [has the `url` field](https://github.com/jsonresume/resume-schema/blob/7095651fbbb593d2c5dc2db3095412b170d74d2e/schema.json#L281-L285).

AmruthPillai:
As much as I love JSON Resume, I find it limiting in certain circumstances. I would still support the freedom to import/export JSON resume schemas, but not to change our schema to theirs. That said, the PR looks great :) Thank you so much for the contribution. Will see to it that it gets released soon.

PR #601, opened 2022-01-10
Added new ContactE without the address block and added it to the Onyx…

### LINKED ISSUE #520: Duplicated address on Onyx layout creates text spillover on top right corner
Have been a fan of Reactive Resume for the past month, but just recently it looks like the Onyx format is duplicating the applicant's address. It normally goes in a designated space below the applicant's name, but now it also appears in the top right column with the other contact information as well. Because addresses are naturally longer than an email or phone number, no matter the font size, the text will spillover halfway through. This looks messy and doesn't need to exist here.

### LINKED ISSUE #548: Repetitive "address" info in Onyx template
Awesome work! Just a little issue regarding the Onyx template: the address information is shown twice, one below the name and a second time at the top right corner. This looks a bit unaesthetic and repetitive. 
Other than that, this one and all your templates are amazing. Keep it up!

mupsys:
@AmruthPillai This is a quick fix to remove the duplicate address. Correct me if I am wrong but it looks like the different Contact components (ContactA, ContactB, etc) seem to have different layouts / styling that go with the appropriate themes? That is the reason that I went this route.

fixes #520, fixes #548

If I can get some time (and of course, you agree), I would like to see if I can come up with a way to add a prop that configures the way the Contact component looks (so instead of having a set of styled Contact components, pass some sort of styling configuration as a prop or something... this will require some more thought and designing). Also, as part of that, I should be able to add a way to customize that block (although the user could always just not put the data in that section I suppose).

### DIFF
--- a/src/templates/Onyx.js
+++ b/src/templates/Onyx.js
@@ -3,7 +3,7 @@ import React, { memo } from 'react';
 import { hasAddress } from '../utils';
 import AwardsA from './blocks/Awards/AwardsA';
 import CertificationsA from './blocks/Certifications/CertificationsA';
-import Contact from './blocks/Contact/ContactA';
+import Contact from './blocks/Contact/ContactE';
 import EducationA from './blocks/Education/EducationA';
 import HeadingA from './blocks/Heading/HeadingA';
 import HobbiesA from './blocks/Hobbies/HobbiesA';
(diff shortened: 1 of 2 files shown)


AmruthPillai:
That's a really good idea, building a single component that changes depending on it's props. I honestly don't know what I was thinking when I built multiple components with different styles like these :D But thank you for your contributions and helping me clean up the messes I've made.

I'm currently at work building the next version of Reactive Resume and I'm taking in all these considerations, and your suggestion has helped a lot. Will keep looking at the issues board for any more lapses in judgement that I can fix in the new one.

mupsys:
No problem. And thank you for bringing this awesome project to existence and making it available for all to use. This is awesome work all around!

PR #609, opened 2022-01-25
add 'object cover' class to templates that have issues with profile image stretching.

apat183:
Image without 'object-class' class on profile picture div in template, stretches.
Screen Shot 2022-01-25 at 8 31 53 PM

Image with 'object-class' class on profile picture div in template. No stretching.
Screen Shot 2022-01-25 at 8 32 42 PM

Only impacting two templates


apat183:
Any chance to get this merge?

PR #669, opened 2022-03-13
Fix-Gengar Theme MastHead Icons Disappear on Theme/Primary Color Change

chandiwalaaadhar:
This is a Fix for a UI Bug which can be Observed in Gengar Template where Icons disappear on Primary Color Change

Follow Steps below to Replicate:

1- Select Template Gengar
2- Fill Email, Mobile details
3- Now Change the Primary Color
4- Keep the Background Color as #ffffff

You will observe that the Icons for Email, Mobile, etc in the MastHead will disappear

Reason: The Icon Color is being overwritten by the Background Color


AmruthPillai (approved):
Looks good to go :)

PR #682, opened 2022-03-14
Fix - Left SideBar Photo Filters Button & Full Name Field Overlapping in iOS Devices

chandiwalaaadhar:
**Issue** - On iOS devices (iPhone, iPad, iPod) it was observed as shown in image below, that the Full Name (ResumeInput Component) and Photo Filter Button were overlapping because the CSS was not behaving as expected.

You will be able to replicate this issue on Safari in Mac by Activating "Enter Responsive Design Mode" or on an iPhone ( Tested and verified on iPhone 11 and 8 Plus )

This fix will be only reflected on the iOS device and _not_ on Safari in Mac for obvious reasons (check isiOS util Function)

Screenshot 2022-03-14 at 11 29 27 AM

### DIFF
--- a/client/components/build/LeftSidebar/sections/Basics.tsx
+++ b/client/components/build/LeftSidebar/sections/Basics.tsx
@@ -5,6 +5,7 @@ import { useState } from 'react';
 
 import Heading from '@/components/shared/Heading';
 import ResumeInput from '@/components/shared/ResumeInput';
+import isiOS from '@/utils/isiOS';
 
 import PhotoFilters from './PhotoFilters';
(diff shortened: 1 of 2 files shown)


AmruthPillai:
I appreciate the continuous contributions you are bringing into Reactive Resume, first off, thank you for that! :)

Specifically for this issue, I feel it can be solved in a non-JS way, but just using `display: grid` instead of `display: flex` and tweaking it as per usage because flex-gap is not supported on webkit/safari yet. I'll handle raising a fix for this, but thank you so much for bringing this to light, and also coming up with a solution for it.

PR #685, opened 2022-03-14
Fix - Work Experience Website Link Redirects to 404 if manually entered without http/https

chandiwalaaadhar:
Issue - When user enters the "Website" in Work Experience and enters it without http or https, on clicking on that link, the user gets routed to 404 Page

This issue exists in all 6 templates currently

Fix- Added addHttp(url) in link prop in Section.tsx files of all the 6 templates


AmruthPillai:
How silly of me, I missed out on adding the utility function in the most important place 😄  Thank you so much for this, and also kudos to finding out about the existing `addHttp` function.

chandiwalaaadhar:
Thanks @AmruthPillai and Kudos to you for writing such a developer friendly & readable code, it wasn't hard to find these utility functions 👍🏻 😀

Issue #702, opened 2022-03-15
Text veering off from Artboard.


RahulWebb:
**Text veering off from Artboard on both the 'preview' and 'downloaded pdf'**

**To Reproduce**
1. Go to 'Certification section'
2. Click on 'website or summary'
3. Stored character count 'more than 80'
4. Scroll down to 'edit'
5. Bug spotted while using "2 columns" setting
6. See error

**Expected behavior**
Clearly, text veering isn't supposed to happen 

**Screenshots**
1
![Screenshot (5)](https://user-images.githubusercontent.com/68050000/158366801-b8fcc9ec-02a9-49c4-bf67-9ec94016945a.png)
 screenshots attached 

**Desktop (please complete the following information):**
 - OS: windows 11
 - Browser: chrome 
 - Version: 99.0.4844.51

**N/A**

**Additional context**

PR #704, opened 2022-03-15
Fix-Bug#702-Text Veering from Artboard (Sidebar)

chandiwalaaadhar:
Issue- #702 (This issue exists on the whole sidebar content, and not limited to certifications)

Fix- I tried using tailwind's "break-words" utility in *template*.module.scss files but it seems to fail when some Website Url is entered (Certificate Section). Therefore using {wordBreak:"break-word"}  in inline styles works as expected. And the issue looks fixed. 

Please suggest if there is a better way.

### DIFF
--- a/client/templates/Castform/Castform.tsx
+++ b/client/templates/Castform/Castform.tsx
@@ -27,7 +27,7 @@ const Castform: React.FC<PageProps> = ({ page }) => {
       <div className={styles.container}>
         <div
           className={clsx(styles.sidebar, css(`svg { color: ${color} } --primary-color: ${color}`))}
-          style={{ color, backgroundColor: theme.primary }}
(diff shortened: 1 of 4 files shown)


AmruthPillai:
This is a good fix, but I'd like to keep as much of the styling away from inline CSS and keep it within the modules. I'm not really sure why it didn't work when you added it into the modules.scss folder. I'll try to take a stab at it as soon as I get the time, but this is great work, and good idea using the `word-break` property at the right moment :)

AmruthPillai:
Okay, resolved this using the same CSS property, but instead of adding it as a inline CSS style, added it within modules. The issue was because tailwind CSS's implementation of break-words is quite different, not sure why.

https://github.com/AmruthPillai/Reactive-Resume/commit/b2f1fb3a5502988a49c5cd3e496d9d165f5c1792

PR #714, opened 2022-03-16
Fix- Publication Modal Label text

chandiwalaaadhar:
Issue- Publication Modal Screen Label Text Incorrect as in the Image Below

Fix- Removed the Double Quotes wrapping the Variable

Screenshot 2022-03-16 at 11 14 58 AM


AmruthPillai:
Good catch! 🎣

PR #720, opened 2022-03-16
Fix- Resume Shared is not Scrollable on Phone Screen

chandiwalaaadhar:
Issue- Resume when shared as Link, and viewed on Mobile Devices is not Scrollable to the left ( towards left sidebar )

Fix- Minor Changes the Preview.module.scss file


AmruthPillai:
Perfecto, I've noticed this as well. Thank you for raising a PR for this :)

Issue #725, opened 2022-03-16
[BUG] Error when changing colour with custom HEX


blacksoulgem95:
**Describe the bug**
The front-end crashes after colour change and the CV is not accessible anymore.

**To Reproduce**
1. Create a resume
2. Paste a HEX Colour into the colour field

**Expected behavior**
The colour to be the one I've chosen 

**Screenshots**

**Desktop (please complete the following information):**
 - OS: [e.g. iOS] MacOS 
 - Browser [e.g. chrome, safari] Chrome
 - Version [e.g. 22] Latest stable as of 16th March 22

**Smartphone (please complete the following information):**

**Additional context**

Affected resume: https://rxresu.me/sofivic95/sofia-vicedomini

AmruthPillai:
May I know which color you tried to enter? There's a hex pattern matcher which accepts any valid hex color. It should accept only and only when it matches. Curious to know what could have broken this check.

blacksoulgem95:
I tried to enter #db2877

blacksoulgem95:
I want to notify you that it still throws errors when re-opening the affected CV.

<img width="1791" alt="Schermata 2022-03-18 alle 14 15 33" src="https://user-images.githubusercontent.com/11537307/159009552-bbd162ae-63c8-4938-9ea7-16bbefde6974.png">

AmruthPillai:
@blacksoulgem95 Your resume should be accessible again now.

blacksoulgem95:
thanks!

PR #727, opened 2022-03-17
Fix-Crash on Entering Primary Color Hex Code without # Prefix

chandiwalaaadhar:
Issue- #725

### DIFF
--- a/client/components/build/RightSidebar/sections/Theme.tsx
+++ b/client/components/build/RightSidebar/sections/Theme.tsx
@@ -19,6 +19,9 @@ const Theme = () => {
   const { background, text, primary } = useAppSelector<ThemeType>((state) => get(state.resume, 'metadata.theme'));
 
   const handleChange = (property: string, color: string) => {
+    if (color[0] !== '#') {
+      color = `#${color}`;
+    }
     dispatch(setResumeState({ path: `metadata.theme.${property}`, value: color }));
   };
 
--- a/client/components/build/RightSidebar/sections/Theme.tsx
+++ b/client/components/build/RightSidebar/sections/Theme.tsx
@@ -19,10 +19,7 @@ const Theme = () => {
   const { background, text, primary } = useAppSelector<ThemeType>((state) => get(state.resume, 'metadata.theme'));
 
   const handleChange = (property: string, color: string) => {
-    if (color[0] !== '#') {
-      color = `#${color}`;
-    }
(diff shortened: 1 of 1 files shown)


AmruthPillai:
Could this probably accept non hexadecimal values as well? Is it safe to assume that the value input it always a valid hex color?

client/components/build/RightSidebar/sections/Theme.tsx
```diff
@@ -19,6 +19,9 @@ const Theme = () => {
   const { background, text, primary } = useAppSelector<ThemeType>((state) => get(state.resume, 'metadata.theme'));
 
   const handleChange = (property: string, color: string) => {
+    if (color[0] !== '#') {
+      color = `#${color}`;
```
amruthpillai: It'll be preferred to create a new temporary color variable and reassign that instead of modifying the prop directly.

chandiwalaaadhar: > Could this probably accept non hexadecimal values as well? Is it safe to assume that the value input it always a valid hex color?

The Current Behaviour is that if the input value is an INVALID hex, the color doesn't change
Tested with Invalid Hex Values like- ```12876( , ------, %*&*$^*,  #12345aa```

chandiwalaaadhar:
> Could this probably accept non hexadecimal values as well? Is it safe to assume that the value input it always a valid hex color?

The Current Behaviour is that if the input value is an INVALID hex, the color doesn't change
Tested with Invalid Hex Values like- ```12876( , ------, %*&*$^*,  #12345aa```

client/components/build/RightSidebar/sections/Theme.tsx
chandiwalaaadhar: Oh! I just read about it, It's bad practice to modify the prop directly. Thanks Amruth!

chandiwalaaadhar:
@AmruthPillai 
[https://github.com/AmruthPillai/Reactive-Resume/pull/727/commits/e4950728d8b4f886e7e9842c9fc3497d1fd938fb](https://github.com/AmruthPillai/Reactive-Resume/pull/727/commits/e4950728d8b4f886e7e9842c9fc3497d1fd938fb)

Would this be fine? It Makes the code look cleaner and no need to assign to any variable.

AmruthPillai:
Looks good :)

PR #738, opened 2022-03-18
Update docusaurus to prevent crash with client

martadinata666:
on `pnpm dev` , docusaurus not conflicted with client because it set port on launch.
`docusaurus start --no-open --port 3200`

But on `pnpm start` client and docusaurus conflict each other, as the command not define port.
`docusaurus serve`

this PR add port to docusaurus, so `pnpm start` can run correctly.
`docusaurus serve --port 3200`


AmruthPillai:
Gotcha, missed that out as I don't use docusauraus serve to host the docs. Thank you for the PR!

Issue #697, opened 2022-03-14
[FEATURE] Docker Healthchecks


modem7:
Container healthchecks for server, client and DB. In case something fails for whatever reason (usually it'd either be the server or the DB), it can be automatically restarted. 

This can further be extended if you so wish by utilising "depends_on" in your compose file.

e.g.
```yaml
depends_on:
  postgres-database:
    condition: service_healthy
```

This can either be done via compose, and/or via the dockerfile itself. 

Server:
```yaml
    healthcheck:
      test: curl -fSs http://localhost:3100/health || exit 1
      interval: 30s
      timeout: 5s
      retries: 3
```

Client:
```yaml
    healthcheck:
      test: curl -fSs 127.0.0.1:3000 || exit 1
      interval: 30s
      timeout: 5s
      retries: 3
```

DB:
```yaml
    healthcheck:
      test: ["CMD-SHELL", "pg_isready -U postgres" ]
      interval: 30s
      timeout: 5s
      retries: 3
```

---
<details>
<summary>Example in use:</summary>

```yaml
 ###################
 ##Reactive Resume##
 ###################

  reactiveresumeserver:
    image: amruthpillai/reactive-resume:server-latest
    container_name: ReactiveResume-Server
    environment:
      - PUBLIC_URL=https://resume.$DOMAINNAME
      - POSTGRES_HOST=reactiveresumedb
      - POSTGRES_DATABASE=$RR_POSTGRES_DATABASE
      - POSTGRES_PASSWORD=$RR_POSTGRES_PASSWORD
      - POSTGRES_USERNAME=$RR_POSTGRES_USERNAME
      - TZ=$TZ
      - SECRET_KEY=$RR_SECRET_KEY
      - JWT_SECRET=$RR_JWT_SECRET
      - JWT_EXPIRY_TIME=604800
      - PUBLIC_GOOGLE_CLIENT_ID=$RR_PUBLIC_GOOGLE_CLIENT_ID
      - GOOGLE_CLIENT_SECRET=$RR_GOOGLE_CLIENT_SECRET
      - GOOGLE_API_KEY=$RR_GOOGLE_API_KEY
    depends_on:
      - reactiveresumedb
    networks:
      pihole:
        ipv4_address: '172.22.0.140'
      isolated:
    labels: 
      - autoheal=true
      - "traefik.enable=true"
      ## HTTP Routers
      - "traefik.http.routers.resumeserver-rtr.entrypoints=https"
[...]

PR #755, opened 2022-03-21
Use Anchor values and Extension Fields instead

Leopere:
Instead of env_file using yaml anchor values and extension fields to allow a more broad set of deployment options.

Needs approval/testing.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,4 +1,35 @@
 version: '3'
+
+x-env: &env
+  environment:
+  # App
+  - TZ=UTC
+  - SECRET_KEY=
+  # URLs
+  - PUBLIC_URL=http://<SERVER-IP>
+  - PUBLIC_SERVER_URL=http://<SERVER-IP>/api
+  # Database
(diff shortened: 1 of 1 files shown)


Leopere:
This configuration allows for easy deployment to a Docker Swarm stack as well as Docker-Compose stack in case the user can't use .env files in their deployment.

AmruthPillai:
I see that almost all of the ENVs are being passed to all of the services. While that was also the case with the .env file, I did that because I wanted to reduce the number of lines of code. But in this case, to maximize efficiency, we could split the x-env variable into 3, for each service, and just pass the env that is required for it. You would be able to hit two proverbial birds with one stone that way (not really of course).

For a breakdown of which service requires which ENV, I've shared that information in this comment:
https://github.com/AmruthPillai/Reactive-Resume/issues/697#issuecomment-1070913218

AmruthPillai:
This has been implemented in the latest tag of [v3.2.10](https://github.com/AmruthPillai/Reactive-Resume/releases/tag/v3.2.10)

PR #811, opened 2022-04-03
A new template - Leafish

klejejs:
Hello!

I love your CV editor, it is awesome! Thank you very much for the work you have put into it.

I actually already have a CV created, but it id doesn't allow any edits, with your editor, however, I can easily add, remove and change things, thus I created a new template that kind of matches my CV.

In the hopes that it can turn out to be useful for other people too, I am making this PR that includes a new template called Leafish. See the example image below.

P.S. If I understood correctly, the templates are named after Pokemons. I am not a huge fan, thus I chose the name that attracted me the most 😄  

image


AmruthPillai:
WOW, I mean, just wow!

I thought for sure I had overcomplicated the template engine to the point where it's going to be difficult for anyone other than me to create new templates, but you proved me wrong. Thank you for this insanely huge contribution to Reactive Resume, and with a great design I might add. It didn't occur to me to use two different headers the way you have done, it looks quite nice.

And great choice on the name too. I don't know about this Pokemon especially, but suits the template quite well. Thank you so much for your contribution, will test, merge and deploy this very soon.

chandiwalaaadhar:
This Looks great! 👍 😄

PR #817, opened 2022-04-07
Automatic multi-platform Docker image build

schklom:
Uses QEMU and Buildx to make the Docker image available on ARM platforms (can be extended to linux/amd64,linux/arm64,linux/riscv64,linux/ppc64le,linux/s390x,linux/386,linux/mips64le,linux/mips64,linux/arm/v7,linux/arm/v6).

ARM is a must for the convenience of Raspberry Pis :)


AmruthPillai:
Would have to revert this as there is no image for arm/v7 on playwright, and as a result the docker builds are failing. :( Will update once there is a update to the image upstream.

PR #819, opened 2022-04-08
combine-tags

martadinata666:
Only build once for every client and server, resulting same image on both repo image.


AmruthPillai:
I was curious how I would be able to do this, thanks for the PR!

PR #906, opened 2022-06-06
Add File based Storage toggle for Photos

dvd741-a:
Add toggle env option to toggle between file based or S3 Bucket storage.


dvd741-a:
@AmruthPillai - Can you have a look at this merge request ... ?

AmruthPillai:
Quite an interesting solution, thank you for your contribution! :)

AmruthPillai (approved):

AmruthPillai:
@dvd741-a Would you also be able to add the local file storage path to docker volumes, so that on every docker restart/reload, the uploaded images are not erased?

dvd741-a:
> @dvd741-a Would you also be able to add the local file storage path to docker volumes, so that on every docker restart/reload, the uploaded images are not erased?

I would be able to as soon as I can test it :-)
Wasn't able to build the docker image myself using the dockerfile in the repository. If a version gets bumped/built and put on dockerhub, I can try the correct path for the mount

michael-robbins:
I believe inside the container `/app/server/dist/assets/uploads` is the path being referred to in this PR.

So adding something like: `VOLUME /app/server/dist/assets/uploads` inside `server/Dockerfile` should do the trick here!

dvd741-a:
@michael-robbins Can you have a look if this seems correct to you? / Not sure who can approve. - I've already mapped the path to a folder in my instance, as soon as a new version is built & gets put on dockerhub - It should start working.

michael-robbins:
It looks fine to me! I've already updated my setup to use the new env vars as well as the volume!

PR #951, opened 2022-07-25
Cleanup superfluous docker-compose.yml declarations

Leopere:
Removed public port bind for Postgresql, which unnecessarily exposes the Postgres container to the internet publicly.

Also, container names are automatically defined by service name so removed static `container_name` declarations.


AmruthPillai:
I agree with having to remove the port so that postgres isn't accessible to the outside world, and will cherry pick that change and push it in the next release. But the reason I added container_names for every container is actually because whenever I try running docker compose on my project, it would prepend the folder name (project name) along with it, and I didn't actually want that to show up.

AmruthPillai:
@Leopere That's interesting, did not know about the limitations with Docker Swarm, but good enough of a reason to convince me. Will include the PR back in again, thank you! :)

Leopere:
Nice!

PR #1023, opened 2022-10-04
Add PDF file caching

klejejs:
Currently, every time someone opens a resume and tries downloading it as a PDF, it takes some time to generate it, which is not a great user experience. This PR adds a PDF file caching feature, thus each generated PDF is kept in the exports folder and recreated only if it has been updated.
To achieve this, I also had to update the sync saga to save the updated resume to the store as `updateAt` date is generated only during calling the `save` function on the server.


AmruthPillai:
@klejejs So, does this mean all resumes (even the out of date ones) would be stored indefinitely on the file system? This would increase the storage usage I believe, and not something planned for.

klejejs:
@AmruthPillai RIght. I did not think about the instance you are hosting and that very many people are using it. Do you have any data about how many people export PDFs daily? Would you accept this PR if I created an environment variable with cached PDF storage time and then, based on it, delete old PDF files?

AmruthPillai:
Yes, as per the function execution log, there's an average of 8000 calls per day. Some days have extreme traffic, and have more than 12000 executions (PDFs generated per day). With each PDF roughly being around 300 KB, that's almost 2.4 GB a day 🤯 

It doesn't necessarily have to be an env var, but a constant in the code to schedule deletion of the file after maybe 6 days? NestJS's scheduler service should be able set this up (similar to how it is now, but just with an increased time) and along with your change on lastUpdate would work well.

klejejs:
Wow, those are some big numbers 🤩 

I'll update the PR then and add the scheduled deletion. I was thinking of an environment variable just because for people like me, who self-host this great piece of software, the deletion period could be set to longer. Also, it will allow you to easily change the period if the exports folder gets too big.

klejejs:
@AmruthPillai Added scheduled deletion and set default timeout to 6 days.

client/store/sagas/sync.ts
```diff
@@ -12,23 +12,24 @@ import {
...
   duplicateItem,
   editItem,
+  setResume,
   setResumeState,
 } from '../resume/resumeSlice';
 
 const DEBOUNCE_WAIT = 1000;
 
-const debouncedSync = debounce((resume: Resume) => updateResume(resume), DEBOUNCE_WAIT);
+const debouncedSync = debounce((resume: Resume, dispatch: AppDispatch) => updateResume(resume).then((resume) => dispatch(setResume(resume))), DEBOUNCE_WAIT);
```
amruthpillai, line 21: @klejejs Just a bit curious about this. Would this not cause side effect changes to the frontend after the fact, for ex. the user types in a bunch of stuff and waits for while, and then continues to type some more details but the promise is already running. This could overwrite the user's current changes, would it not? It's a very very tight window, I'm sure, but more noticeable on slower internet speeds.

What would be the issue if we didn't update state in the sync fn?

server/src/config/config.module.ts
```diff
@@ -52,12 +53,15 @@ const validationSchema = Joi.object({
   STORAGE_URL_PREFIX: Joi.string().allow(''),
   STORAGE_ACCESS_KEY: Joi.string().allow(''),
   STORAGE_SECRET_KEY: Joi.string().allow(''),
+
+  // Cache
+  PDF_DELETION_TIME: Joi.number().default(6 * 24 * 60 * 60 * 1000), // 6 days
```
amruthpillai, line 58: @klejejs +1 on the separate namespace, that shows true experience 💯

client/store/sagas/sync.ts
klejejs: The issue is that without any database updates, the `lastChanged` date field in DB does not update, thus the PDF generator cannot check if the existing file is up-to-date with data or not.

I could investigate sending only new changes to the server to decrease request size but I believe that could be a lot of work.
Another option would be to detect changes made to UI after the request and not run `dispatch(setResume(resume))` when the request completes if the resume version is newer than the one saved in DB. Of course, then special handling would be needed for cases when the request fails and there should also be some checks to make sure the user does not lose the changes when they suddenly decide to close the page.

What do you think? Which would be the better option?

Issue #1024, opened 2022-10-07
[BUG] On Clicking enter while filling the form, entire website is getting re-loaded.


jaichaturvedi88:
**Describe the bug**

On Clicking enter while filling the form the page is getting re-loaded.

**Product Flavor**

- [ ] Managed (https://rxresu.me)

**To Reproduce**

1. Open any form to enter data. Let's say interests form. 
2. Type the name of interest
3. Press enter
4. Entire page will be reloaded.

**Expected behavior**

After pressing enter post filling the form, the data should be submitted, instead the page is getting reloaded.

kmkhant:
Hi there, I just fixed the Interest Form locally, I will submit pull request soon. Thanks for the info

PR #1032, opened 2022-10-10
FIX whole page reload when press enter

kmkhant:
fix [#1024](https://github.com/AmruthPillai/Reactive-Resume/issues/1024#issue-1400709220)

This is my first contribution to an open source project. Please guide me if I were wrong.


AmruthPillai:
That's a good fix, does this occur only in the InterestModal, or is this something we might need in other modals/forms as well?

kmkhant:
As far as I tested, I found page reloading issue when pressed enter only in Interest Modal. But it would be a great user experience if the user can just press enter after filling all the required forms instead of clicking the button.

AmruthPillai:
Let's do it for all the modals, it would be a great help :) Thank you!

kmkhant:
OK, I will update my pr soon thanks for the suggestion <3

AmruthPillai:
Thank you so much for the contribution, will make sure it goes out on the next update! 🔥

kmkhant:
Thank you so much too for creating this great app. I've learned a lot by reading the source code. I'll try to make more contributions in the future.

PR #1058, opened 2022-10-25
Fix server crash when non-zip file is uploaded

### LINKED ISSUE #1048: [BUG] Uploading a non-ZIP file causes server to crash
**Whenever I upload a non-zip files in the LinkedIn import section, it crashes the server**

**Product Flavor**

- [x] Managed (https://rxresu.me)
- [x] Self Hosted

**To Reproduce**

1. Go to 'dashboard'
2. Click on 'Import From External Sources'
3. In Import from LinkedIn section, click Upload ZIP Archive button.
4. Upload a file that is not a ZIP
5. See the error in the logs
6. If you reload, the server throws a 504.

**Expected behavior**

Give us an error that the file type is incorrect.

**Desktop (please complete the following information):**

- OS: Windows 10 21H2 Build 19044
- Browser Chrome
- Version 3.6.8+1

klejejs:
This fixes #1048.
Turns out the issue was that though an invalid archive is not empty, it cannot be closed and this was crashing the server.

### DIFF
--- a/server/src/integrations/integrations.service.ts
+++ b/server/src/integrations/integrations.service.ts
@@ -35,6 +35,7 @@ export class IntegrationsService {
 
   async linkedIn(userId: number, path: string): Promise<ResumeEntity> {
     let archive: StreamZip.StreamZipAsync;
+    let isArchiveValid = false;
 
     try {
       archive = new StreamZip.async({ file: path });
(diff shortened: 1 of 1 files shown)


AmruthPillai:
Great catch, didn't think to realize that could have been the issue. Thank you so much for this!

PR #1060, opened 2022-10-27
feat: additional work sections

sashokbg:
I would like to open this PR as a discussion on how to better arrange work experiences along multiple pages.

I have added a button "Add new section" under the current "work" section that adds a work-1, work-2 and so on sections.
I have modified the WorkModal to work with changing path (similar to the custom sections).

Please let me know what you think.

https://gfycat.com/thesesecretamericanratsnake

### DIFF
--- a/client/components/build/LeftSidebar/sections/Section.tsx
+++ b/client/components/build/LeftSidebar/sections/Section.tsx
@@ -1,6 +1,6 @@
 import { Add } from '@mui/icons-material';
 import { Button } from '@mui/material';
-import { ListItem } from '@reactive-resume/schema';
+import { ListItem, Section as SectionRecord, SectionType } from '@reactive-resume/schema';
 import clsx from 'clsx';
 import get from 'lodash/get';
 import { useTranslation } from 'next-i18next';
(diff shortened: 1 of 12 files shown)


AmruthPillai:
This is an interesting idea, and thank you very much for the screen cap demonstrating your feature implementation. That saves a lot of time.

I am just curious about the naming. "Add New Section" is a bit ambiguous as there's already a "Add New Section" in the footer. Would "Duplicate Section" be more appropriate? (this also has the confusion that items in the section would be copied over). Any other suggestions come to mind?

Once again, thank you so much for your contributions to Reactive Resume <3

sashokbg:
Hello @AmruthPillai,

Yes you are right I will change the message.
I was thinking that any section should in theory be duplicate-able.

We should also make sure that users can clearly understand that duplicating a section is so that it can be arranged in the layout on different pages. Maybe a on-hover tooltip ? What do you think ?

Best Regards and thank you for the motivating message !
Alex

sashokbg:
Hello @AmruthPillai I have cleaned the code a bit and rendered it more generic. I also added the label "Duplicate Section" in English only so far.
Please let me know what you think.

client/components/build/LeftSidebar/sections/Section.tsx
```diff
@@ -10,28 +10,34 @@ import Heading from '@/components/shared/Heading';
...
 import { useAppDispatch, useAppSelector } from '@/store/hooks';
 import { ModalName, setModalState } from '@/store/modal/modalSlice';
-import { duplicateItem } from '@/store/resume/resumeSlice';
+import { duplicateItem, duplicateSection } from '@/store/resume/resumeSlice';
 
 import SectionSettings from './SectionSettings';
 
 type Props = {
   path: `sections.${string}`;
+  type?: SectionType;
```
sashokbg, line 19: I needed to add type to sections so that we can have multiple work sections that share the same modal by type but have different paths

client/components/build/LeftSidebar/sections/Section.tsx
```diff
@@ -42,21 +48,42 @@ const Section: React.FC<Props> = ({
 
   const handleAdd = () => {
     const id = path.split('.')[1];
-    const modal: ModalName = validate(id) ? 'builder.sections.custom' : `builder.${path}`;
+    let modal: ModalName = validate(id) ? 'builder.sections.custom' : `builder.${path}`;
 
+    if (type) {
+      modal = `builder.sections.${type}`;
```
sashokbg, line 54: Find modal by type instead of path. This way we can have one modal "**builder.sections.work**" that works with paths "**builder.sections.work-1, work-2**" and so on

AmruthPillai:
@sashokbg I'll have a look over this PR and the other one on the weekend as I'm tied up with work right now. Just letting you know if you're wondering what's taking so long :D But thank you so much for your amazing contribution to the project. It really means a lot, to me and the community.

PR #1069, opened 2022-11-06
fixed grammatical and stylistic errors in Hungarian translation

arvaid:
The original translation was machine-generated, with many grammatical and stylistic errors that showed up not just in the application but in the resume content as well.


AmruthPillai:
Thank you for raising the PR. I can't really verify every contribution that comes in, but I'm thankful for folks like you to come in and correct them when you see something wrong.

PR #1107, opened 2022-11-25
fix: made some missing texts translatable

tryallthethings:
BREAKING CHANGE: locales without the new fields will display the field name

### DIFF
--- a/client/components/build/RightSidebar/sections/Layout.tsx
+++ b/client/components/build/RightSidebar/sections/Layout.tsx
@@ -103,7 +103,7 @@ const Layout = () => {
                   <Droppable key={index} droppableId={index}>
                     {(provided) => (
                       <div ref={provided.innerRef} className={styles.column} {...provided.droppableProps}>
-                        <p className={styles.heading}>{columnIndex ? 'Sidebar' : 'Main'}</p>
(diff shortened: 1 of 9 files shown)


AmruthPillai:
@tryallthethings Wow, vielen dank! This will require some time to get to production, as I'd need some help from translators to fill in the other fields/placeholders as well, but nevertheless, will ensure that this goes out soon :) 

As for the Language Name, I think it's okay to skip the "Formal / Sie" as it can be implied, as there's no other variations of German that's being offered.

I hope the keys were easy enough to translate to, considering how complicated the language can get.

AmruthPillai:
Oh, and if you're ever in Berlin some time, I owe you a beer! 🍻

tryallthethings:
> This will require some time to get to production, as I'd need some help from translators to fill in the other fields/placeholders as well, but nevertheless, will ensure that this goes out soon :)

Sure, take your time. 😄 

> As for the Language Name, I think it's okay to skip the "Formal / Sie" as it can be implied, as there's no other variations of German that's being offered.

I was initially planning to add an informal translation as well. But as this will be mostly used in the professional space, I agree that this can be omitted. And from what I can see so far, this is only relevant on the backend anyway.

> I hope the keys were easy enough to translate to, considering how complicated the language can get.

I spent way too much time getting the development environment running. After that, it wasn't too complicated to identify the files and lines to modify - even without any knowledge of Typescript.

> Oh, and if you're ever in Berlin some time, I owe you a beer! 🍻

I don't have any plans to come to Berlin anytime soon. But thanks for the offer! 🍻

PR #1114, opened 2022-11-26
Fix for links in PDFs, template fix

tryallthethings:


client/templates/shared/DataDisplay.tsx
```diff
@@ -15,7 +15,7 @@ const DataDisplay: React.FC<React.PropsWithChildren<Props>> = ({ icon, link, cla
     return (
       <div className={clsx('inline-flex items-center gap-1', className)}>
         {icon}
-        <a href={link} target="_blank" rel="noreferrer" className={textClassName}>
+        <a href={link} target="_blank" rel="noreferrer" className="{textClassName}">
```
amruthpillai, line 18: But this wouldn't actually apply the class to the element, right?

tryallthethings: 🤦‍♂️ Yes, you're right. Didn't double-check the HTML after adding the quotation.
```html
<a href="https://linkedin.com/tester" target="_blank" rel="noreferrer" class="{textClassName}">tester</a>
```
But that was the only solution I was able to find besides removing `textClassName` completely. Maybe you got at least a starting point to look into this 😄

PR #1155, opened 2022-12-29
fix: overflow-y-auto

coolswood:
If "Show scroll bars" is set to "Always" in the mac settings. Sidebars scroll horizontally all the time. Setting overflow: auto fixes the bug
Screenshot 2022-12-29 at 19 41 22


AmruthPillai:
Interesting, thank you for the PR. I always keep my scrollbars hidden (even when a mouse is connected) so seemed to have missed this.

PR #1204, opened 2023-01-28
Removed unnecessary workspaces declaration from package.json

mbsc26:
Sir,

Workspaces declaration is not required in the root "package.json" in case you are using `pnpm` as the package manager for your monorepo. Workspaces declaration is only required inside `pnpm-workspace.yaml` which is already there.

We declare workspaces in "package.json" only if we are using `npm` as the package manager for our monorepo.

I have thus removed it from the root "package.json". Kindly have a look.

Reference:- https://turbo.build/repo/docs/handbook/workspaces

Wishing you success,

Muhammad

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,11 +11,6 @@
     "format": "prettier --write .",
     "lint": "turbo run lint"
   },
-  "workspaces": [
-    "schema",
-    "client",
-    "server"
-  ],
   "dependencies": {
     "cross-env": "^7.0.3",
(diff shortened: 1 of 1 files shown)


AmruthPillai:
I added the workspaces key so that anyone who is trying to build the app locally is not restricted to use `pnpm`. They could use npm or Yarn if they wish, with a bit more set up, but the `workspaces` property inside package.json stands as a reference for other package managers.

Hope that clears it up! :)

PR #1225, opened 2023-02-09
Translation fixes on LeftSideBar

raphaelcangucu:
- Adding pt-BR translation file
- Fixing all sections translate that wasn't working
- Fix formatDate error that was subtracting one day from the date


AmruthPillai:
Thank you very much for the language additions and changes. It is very much appreciated. Thank you for the support to keep Reactive Resume going.

raphaelcangucu:
Hi Amruth,

What do you think about create a oauth server for this project?

I am looking for a job and my main pain is to put in every site the same
information.

That’s why I searched for a solution similar to Reactive Resume.

So, the oauth server will share this information with any site that
implement the login with reactive resume.

Kind regards

Em sex., 10 de fev. de 2023 às 12:04 AM, Amruth Pillai <
***@***.***> escreveu:

PR #1271, opened 2023-03-29
feat: Updated landing page layout

JacksonTyrannosaurusTech:
image


AmruthPillai:
It's quite good, a lot cleaner than the original layout of course.

But wouldn't it be good to have some insight into the application without forcing the user to sign up? The screenshots section was meant to do just that, so they could sneak a peek at how it looks without having to sign up.

Issue #1302, opened 2023-04-30
JSON Uploading


Bhaweshgoyal:
**Describe the bug**

I have downloaded the JSON file and uploaded the same file again without editing it not giving proper details after uploading it changes the, template, font color and theme 
And mainly not working after upload of JSON file shows only top navBar details like name , heading description and all

**Product Flavor**

- [x] Managed (https://rxresu.me)
- [ ] Self Hosted

**To Reproduce**
Download the fully created Resume and download its JSON file and re-upload it  get reproduced

1. Go to download JSON file
2. Click on Instant upload 
3. upload your JSON file

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**

- OS: 
- Browser 
- Version 

**Additional context**

AmruthPillai:
This issue was automatically closed as it was created before 2024-01-01.

Issue #1296, opened 2023-04-27
[FEATURE] Generate resume through API


mael-belval:
**Is your feature request related to a problem? Please describe.**

It can be useful to be able to generate a resume trough an API call. This way, we would be able to generate a resume without having to register.

**Describe the solution you'd like**

An new route returning a PDF based on data provided in JSON. Profile picture would be provided as external link.

**Additional context**

Currently, you must create an account to create your resume and fill the required informations.

AmruthPillai:
This issue was automatically closed as it was created before 2024-01-01.

PR #1306, opened 2023-05-01
Feature/allow to create resume in api

mael-belval:
# Issues solved

#1296 #1302 

# Summary

Allow to generate a PDF resume with an API call on the frontend client.

`POST /api/generate_resume body:{ reactiveResumeV2Json }`

# Details

This new route call `/printer` on the server, cache the JSON based on the 10 first characters of the MD5 hash, build an according PDF and return the public URL. 
The caching is required for optimisation and providing the resume schema to the `/:username/:slug/printer` page of the client app to build the according resume.

# Technical choices

To handle the username, a new configuration key is added, `JSON_CACHE_CLIENT`, which disallow user to use the chosen username (default is `__from_json__`).
PDF are exported in a new folder, `assets/exportsFromJson`.
To validate the sent json, the library zod is added to the schema library. Some of the existing types have been converted to zod schema (while preserving the existing type through inference).
This validation of the json schema allowed to fix an existing bug where importing a reactive resume v2 json in the app resulted in data loss.


AmruthPillai:
Not looking into this at the moment as creating a public API opens up some security concerns, and I do not have the time to properly release this as a full-fledged feature. Apologies, but thank you for the contribution and support :)

PR #1411, opened 2023-07-31
fix: Work Modal Datepicker popup

Jubair70:
Expected behavior: In the work modal, the user should be able to select a year, then a month, and then a day. Also should change the date through the keyboard.

Found behavior: After selecting the year it breaks the UI, renders the calendar in the corner of the screen instead of rendering with respect to the Date Range element and the user can't select the follow-up month and date properly.
Also when changing the date by arrow key, it stopped working after the first time and remove the focus from the element.

Screenshot 2023-07-31 at 3 05 54 PM


AmruthPillai:
Thank you for your contribution! :)

Issue #1460, opened 2023-09-08
[BUG] Download wrong PDF


anasbn3issa:
### Is there an existing issue for this?

- [X] Yes, I have searched the existing issues

### Product Variant

Cloud (http://rxresu.me)

### Current Behavior

initially have a profile picture, 
Removes picture from CV 
Export to pdf 
downloaded version -> contains picture still 
only updates the CV version when refreshes page. 

### What browsers are you seeing the problem on?

Chrome

AmruthPillai:
This issue was automatically closed as it was created before 2024-01-01.

PR #1469, opened 2023-09-17
fix: update the `updatedAt` field while deleting/uploading images

marzooqahmed944:
This pull request addresses problem #1460 by modifying the `updatedAt` value within the resume table. It now gets updated whenever a user deletes or uploads their resume photo. Additionally, the front-end also synchronizes with this change. Consequently, whenever a change in the image occurs, the `lastUpdatedAt` query parameter in the API is automatically adjusted, triggering the generation of a new PDF download.


AmruthPillai:
Thank you for looking into this fix, will ensure it goes out in the next release :)

Issue #1458, opened 2023-09-07
[Feature] Break Line for Letter-Sized Page


brycekinsey:
### Is there an existing issue for this?

- [X] Yes, I have searched the existing issues and it doesn't exist

### Feature Description

Hello! This is a great application that actually helped me land my current job after being layed off due to workforce reductions. Thank you so much for all your work @AmruthPillai ! 

In the Theme options, there's the ability to enable a Break Line to show a line on all pages to mark the height of an A4 page. Is it possible to also add this ability for a Letter-sized page? 

![image](https://github.com/AmruthPillai/Reactive-Resume/assets/96494873/e9dbbdeb-077d-44d5-aa60-af6a18263757)

AmruthPillai:
This issue was automatically closed as it was created before 2024-01-01.

PR #1484, opened 2023-09-26
Update break line description when page size is changed

MrRamPoudel:
This pull request addresses the feature request mentioned in #1458 by improving the description of page breaks. Previously, page breaks were displayed correctly for both Letter size and A4 size, but this information was not reflected to the user when a different page format was selected. This update only fixes it for English language.

Changes in this pull request:
- Improved the description of page breaks to indicate the size of the page (Letter or A4).


AmruthPillai:
Thanks, this is great! :)

Issue #1501, opened 2023-10-15
[Bug] the tooltip of the button in the left sidebar is still in English when lang=zh


dakele123:
### Is there an existing issue for this?

- [ ] Yes, I have searched the existing issues

### Product Variant

Self-Hosted

### Current Behavior

<img width="189" alt="image" src="https://github.com/AmruthPillai/Reactive-Resume/assets/14195879/81ef5906-65ec-4f78-99e8-8dedf75a0692"> 
tooltip still in english when choosing zh

### Expected Behavior

<img width="99" alt="image" src="https://github.com/AmruthPillai/Reactive-Resume/assets/14195879/b3033de0-df06-43f3-922f-172cfd33a787">

### Steps To Reproduce

1. swtich lang=zh
2. create new resume
3. Hover the mouse over the 'work experience' column

### What browsers are you seeing the problem on?

Chrome

AmruthPillai:
This issue was automatically closed as it was created before 2024-01-01.

PR #1502, opened 2023-10-15
[fix] the tooltip of the button in the left sidebar is still in Engli…

dakele123:
…sh when lang=zh

#1501


AmruthPillai:
Thank you for your contributions! :)

Issue #1431, opened 2023-08-15
[Bug] PDF download does not work locally with docker compose


esep:
### Is there an existing issue for this?

- [X] Yes, I have searched the existing issues

### Product Variant

Self-Hosted

### Current Behavior

Cannot download resume as PDF in a docker compose setup: "Failed to load resource: the server responded with a status of 504"

### Expected Behavior

Resume is downloaded as a PDF file

### Steps To Reproduce

1. Clone repo
2. Start with docker compose up
3. Open browser (e.g. Chrome) to localhost:3000
4. Sign up a test user
5. Create a resume with any data
6. Download PDF
7. Get the following error in the browser developer console and red cross popup on the lower right without text `:3100/printer/testi/test?lastUpdated=1692103045:1     Failed to load resource: the server responded with a status of 504 (Gateway Timeout)`
8. Downloading does not start

### What browsers are you seeing the problem on?

Chrome

### Anything else?

![RxResume pdf error](https://github.com/AmruthPillai/Reactive-Resume/assets/15034165/b645bb89-cb99-412e-9bc9-632ba46af218)

NateTheSage:
Same here, I just had to reinstall from scratch recently and it fails almost immediately with 504.


![chrome_MMEY3CW8DZ](https://github.com/AmruthPillai/Reactive-Resume/assets/42617513/eb5a3346-67b8-460f-8c0f-1c52e608ed83)

I think I had it working a while ago, but I can't recall.

I'm using a docker-compose file which includes Traefik:
```
version: "3.8"

services:
  postgres:
    image: postgres:alpine
    container_name: rxresume-db
    restart: always
    expose:
      - 5432
    volumes:
      - ${DOCKER_VOLUME_STORAGE:-/mnt/docker-volumes}/rxresume/postgresql:/var/lib/postgresql/data
#      - pgdata:/var/lib/postgresql/data
    healthcheck:
      test: ["CMD-SHELL", "pg_isready -U postgres"]
      start_period: 15s
      interval: 30s
      timeout: 30s
      retries: 3
    environment:
      - POSTGRES_DB=postgres
      - POSTGRES_USER=postgres
      - POSTGRES_PASSWORD=postgres
    networks:
      - proxy

  server:
    image: amruthpillai/reactive-resume:server-latest
    container_name: rxresume-server
    restart: always
    extra_hosts:
[...]

PR #1513, opened 2023-11-04
Fix for #1431

otterDeveloper:
A partial fix for #1431, addresses some of the issues I found in #1431

I implemented url encoding for the secret key that `printer.service.ts` supplies to the Client when generating a pdf version of a resume.

I generated my secret key using `openssl rand -base64 32` which appends an = at the end.
Anyone else also using special characters would also encounter this issue.

The secret key not being encoded caused the Client to not accept it, therefore it refused to display a private resume which ultimately caused a 504 Gateway timeout error.

### DIFF
--- a/client/services/resume.ts
+++ b/client/services/resume.ts
@@ -64,7 +64,7 @@ export const fetchResumeByIdentifier = async ({
 }: FetchResumeByIdentifierParams) => {
   if (!isBrowser) {
     const serverUrl = env('SERVER_URL');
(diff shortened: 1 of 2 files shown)


otterDeveloper:
Also had to implement the url encoding on `client\services\resume.ts`

AmruthPillai:
Thank you for your contributions! :)

cacpmw:
is this released to docker image yet?

PR #1516, opened 2023-11-05
Fix: Bumping next to 13.5.0 broke pnpm install

otterDeveloper:
The recent bump by dependabot of next to 13.5.0 caused `pnpm install` to fail with:

```bash
WARN  GET https://registry.npmjs.org/next/-/next-13.5.0.tgz error (ERR_PNPM_FETCH_404). Will retry in 10 seconds. 2 retries left.
```

It seems `13.5.0` is kinda of an empty / ghost version https://www.npmjs.com/package/next/v/13.5.0
It doesn't even have a readme and npm says it has 0 installs.

Solution:
Bumping next to 13.5.6 fixes the issue.

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -33,7 +33,7 @@
     "md5-hex": "^4.0.0",
     "monaco-editor": "^0.41.0",
     "nanoid": "^3.3.6",
-    "next": "13.4.13",
+    "next": "13.5.6",
     "next-i18next": "^14.0.0",
     "react": "^18.2.0",
     "react-colorful": "^5.6.1",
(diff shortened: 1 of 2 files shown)


AmruthPillai:
Thanks for the catch!

The newer v4 version doesn't even depend on next.js, so I'm quite happy to be moving away from this.

On a side note, I've noticed that you are an active contributor to the project. Firstly, you have my unending gratitude, but also, I was wondering if you'd be interesting in taking some time to alpha test the new version (let me know how easy/difficult it is to self-host, or how the app feels overall). Please reach out to me on hello@amruthpillai.com if that sounds interesting to you! :)

PR #1548, opened 2023-11-20
[v4]docs: Add example of oauth callback url

otterDeveloper:
Add examples for `GITHUB_CALLBACK_URL` and `GOOGLE_CALLBACK_URL`

Took me a while to figure them out.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -61,12 +61,12 @@ REDIS_URL=redis://default:password@localhost:6379
 CROWDIN_PROJECT_ID=
 CROWDIN_ACCESS_TOKEN=
 
-# GitHub (OAuth, Optional)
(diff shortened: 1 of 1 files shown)


AmruthPillai:
That's perfect, thank you for the PR.

Also, I am looking into the possibility of disabling password auth, or other OAuth buttons in case the envs are not set. WebAuthn however, while it sounds exciting, I'm not sure how many people actually use it. I'll keep it on the roadmap however, at least to learn how the WebAuthn APIs work and would be a fun experiment.

otterDeveloper:
I agree Webauthn is not really a priority, just hoping it actually takes off, could make things much better.

PR #1611, opened 2023-11-27
Added a simple python script that can upgrade a V3 JSON file to a V4 JSON file

andy-vdg:
This script will correct dates as well as convert markdown images and links to HTML for the new Rich Text Editor. 
It does not clean up formatting like italic, bold, underline etc. 
Please test it as I am not super familiar with the format the Rich Text Editor expects.


AmruthPillai:
Have you explored using `marked` or some other equivalent library in PyPi to handle the markdown to HTML so it covers all cases, instead of having to do the conversion yourself?

andy-vdg:
Yes - you could use python-markdown for that as well. However when I looked at the HTML tags created by the Rich Text Editor they are not quite standard tags so I went for the quick fix alternative.

Issue #1604, opened 2023-11-27
[Bug] LinkedIn import fails


thepaperpilot:
### Is there an existing issue for this?

- [X] Yes, I have searched the existing issues and none of them match my problem.

### Product Variant

Self-Hosted

### Current Behavior

I exported my LinkedIn data (all of it) and tried importing it, and got a "Oops, the server returned an error." toast. I got the same on the cloud app as well. Best I can tell, based off the docs, I think the issue would be my zip file not having a Certifications or Projects csv file. I don't see any errors in the console nor the docker container's logs.

### What browsers are you seeing the problem on?

Chrome

### What template are you using?

None

CartCaved:
I see the same issue using Firefox.

Pressing Validate gives no error, but when pressing import, the oops message is shows.

I added the two missing sections (projects and certificates) in my linkedin profile and requested a new archive. This time I got an error at validation due to the certificate not having a valid URL (this is not a required field in LinkedIn).

I added a URL in the csv and tried again. At this point the same error as above occured, validation passes but import fails with an oops message.

AmruthPillai:
Even if the zip archive doesn't have one of those files, it should work. I'm surprised that there's an error being thrown because of this.

The issue here is I didn't really test it out with any other LinkedIn Export except my own, so I don't know all the edge cases that could occur.

If you are okay with sharing your data with me, please send your LinkedIn data export zip archive (as-is, with no modifications) to my email hello@amruthpillai.com and I'll take a look asap.

AndreaPi:
Hi @AmruthPillai, I'm having the same issue as @thepaperpilot

jonwtech:
Hello, I'm also experiencing this error. I am a bit stuck since the migration to v4 since the custom sections of my existing resume have disappeared (https://github.com/AmruthPillai/Reactive-Resume/issues/1613) and I can't start again from a fresh import due to this issue.

Snapicek:
Same issue - try on https://rxresu.me/ - firstly there came error of file Certification.csv - there were invalid url or no url - removed manually. Then upload again, it was validated. After clicking on Import - the error of "Ooops, the server returned an error."

DM1122:
> I was having a similar issue and believe it could be due to missing values in the csv file(s).
> 
> In my `profile.csv` file, the second line had something like this:
> 
> ```
[...]

PR #1714, opened 2024-01-12
Update profile.ts to add default values

calumapplepie:
Add some default values to the LinkedIn data import in case of missing data. Default values used in place of optional data in order to minimize changes to other parts of the codebase; since optionals require an extra step to handle.  Default values of whitespace might work better, but I cannot test to be sure.

Closes: #1604


AmruthPillai:
This might help, definitely. I would have preferred them to be empty strings instead of actual values because the end user might get confused as to where this data is coming from. But that's a minor change. Thank you for the contribution! :)

Issue #1726, opened 2024-01-21
[Bug?] After Adding a Link to Text, Subsequent Text Also Has The Link


CorreyL:
### Is there an existing issue for this?

- [X] Yes, I have searched the existing issues and none of them match my problem.

### Product Variant

Cloud (http://rxresu.me)

### Current Behavior

Disclaimer: This is more of a text editor behavior consistency question/discussion, not necessarily a bug!

In other Text Editing software that allow adding hyperlinks to text, after the link has been applied, subsequent text that is typed will not also have the link applied to it

### Microsoft Word

![microsoft-word](https://github.com/AmruthPillai/Reactive-Resume/assets/16601729/3f9f370d-81bf-4256-9a36-5a8371cbca77)

### Google Docs

![google-docs](https://github.com/AmruthPillai/Reactive-Resume/assets/16601729/31235a9e-9c3a-4d78-bdea-9d6950060a11)

However, in Reactive Resume (appreciating that it's from the default behavior from the third-party library [Tiptap](https://tiptap.dev/)), the link will continue to apply to subsequent text that is typed:

### Reactive Resume

![reactive-resume](https://github.com/AmruthPillai/Reactive-Resume/assets/16601729/60d51cb0-010a-4609-af15-44e63c6e7048)

### Expected Behavior

I would expect behavior similar to Microsoft Word and Google Drive, following the design/usability heuristic of ["Consistency and Standards" in the Nielsen Norman Group Top 10 list](https://www.nngroup.com/articles/ten-usability-heuristics/#:~:text=%234%3A%20Consistency%20and%20standards):

### Local Build of Reactive Resume

![reactive-resume-change](https://github.com/AmruthPillai/Reactive-Resume/assets/16601729/8615eb4f-ff0d-43e6-83ed-5248c02b5d7f)

### Steps To Reproduce

1. Open a `Dialog` component that allows the insertion of a `hyperlink` to the `Summary` section of the `Dialog`
2. Type text that you intend to add a hyperlink to
3. Highlight the text made in step 2
4. Click the `Hyperlink` button and add a URL in the `prompt`
5. Continue to type, and observe how the `Hyperlink` continues to apply to subsequently typed text.

### What browsers are you seeing the problem on?

Microsoft Edge

### What template are you using?

None

PR #1727, opened 2024-01-21
[Bugfix] Instantiate Link with inclusive set to false

CorreyL:
If the maintainers of Reactive Resume agree with the points raised in my issue here:

https://github.com/AmruthPillai/Reactive-Resume/issues/1726

This PR will configure [Tiptap](https://tiptap.dev/) to behave similar to other text editors when adding hyperlinks.

### Before

reactive-resume

### After

reactive-resume-change

---

## ⚠️Possible Undesired Side Effect

I will note that this may negatively impact users who wish to add text _after_ adding a hyperlink to the editor:

### Before

link-after-text before

### After

link-after-text after

And as such, I will understand if the Reactive Resume maintainers see this as an overall negative change.

### DIFF
--- a/libs/ui/src/components/rich-input.tsx
+++ b/libs/ui/src/components/rich-input.tsx
@@ -440,7 +440,7 @@ export const RichInput = forwardRef<Editor, RichInputProps>(
         Image,
         Underline,
         Highlight,
-        Link.configure({ openOnClick: false }),
+        Link.extend({ inclusive: false }).configure({ openOnClick: false }),
         TextAlign.configure({ types: ["heading", "paragraph"] }),
       ],
       editorProps: {


CorreyL:
To avoid the undesired side effect mentioned above, perhaps I could extend this implementation to instantiate a new component?

Rather than the simple `window.prompt` here:

https://github.com/AmruthPillai/Reactive-Resume/blob/9298a7473e9f1cf9e87792c872c0511ff6d3f8a8/libs/ui/src/components/rich-input.tsx#L116-L119

I could implement a new `modal` similar to other text editors, where there is a second `input` element that allows for the intended text to be added as well:

### Microsoft Word

image

### Google Docs

image

Issue #1728, opened 2024-01-21
[Feature] Cmd/Ctrl-K Keyboard Shortcut for Adding a Hyperlink


CorreyL:
### Is there an existing issue for this feature?

- [X] Yes, I have searched the existing issues and it doesn't exist.

### Feature Description

The Rich Text Editor already has common keyboard shortcuts such as:

- `Mod-B`: **Bold**
- `Mod-I`: _Italics_
- `Mod-U`: <ins>Underline</ins>

(Where `Mod` is the term used to refer to `cmd` for `Mac` devices, and `Ctrl` for all others in the [Tiptap documentation](https://tiptap.dev/docs/editor/api/keyboard-shortcuts#overwrite-keyboard-shortcuts))

Support for `Mod-K` for add a [Hyperlink](https://rxresu.me/) would help with consistency of users who use common text editor shortcuts.

PR #1729, opened 2024-01-21
[Feature] Keyboard Shortcut for Hyperlink

CorreyL:
https://github.com/AmruthPillai/Reactive-Resume/issues/1728

The Rich Text Editor already has common keyboard shortcuts such as:

- `Mod-B`: **Bold**
- `Mod-I`: _Italics_
- `Mod-U`: <ins>Underline</ins>

(Where `Mod` is the term used to refer to `cmd` for `Mac` devices, and `Ctrl` for all others in the [Tiptap documentation](https://tiptap.dev/docs/editor/api/keyboard-shortcuts#overwrite-keyboard-shortcuts))

This PR adds `Mod-K` for adding a [hyperlink](https://rxresu.me/) when using the Rich Text Editor

hyperlink keyboard-shortcut


libs/ui/src/components/rich-input.tsx
```diff
@@ -113,25 +113,29 @@ const InsertImageForm = ({ onInsert }: InsertImageProps) => {
   );
 };
 
-const Toolbar = ({ editor }: { editor: Editor }) => {
-  const setLink = useCallback(() => {
-    const previousUrl = editor.getAttributes("link").href;
-    const url = window.prompt("URL", previousUrl);
+const setLinkGlobal = (editor: Editor) => {
```
CorreyL, line 116: I'm open to changing the naming of this function if desired, I just decided to leave `Toolbar.setLink` as-is to keep the `diff` of this PR low for now.

libs/ui/src/components/rich-input.tsx
```diff
@@ -440,7 +444,16 @@ export const RichInput = forwardRef<Editor, RichInputProps>(
         Image,
         Underline,
         Highlight,
-        Link.configure({ openOnClick: false }),
+        Link.extend({
+          addKeyboardShortcuts() {
+            return {
+              "Mod-k": () => {
+                setLinkGlobal(this.editor as Editor);
+                return true;
```
CorreyL, line 453: `addKeyboardShortcuts` expects functions in this object to return a `boolean`

image

image

libs/ui/src/components/rich-input.tsx
amruthpillai: I actually don't mind if the code is refactored to be inside Toolbar, and within a useCallback. Would be neater that way. I can take it from here and make the necessary changes.

Thank you for this quick helpful fix.

Issue #1750, opened 2024-02-01
[Bug] Downloaded PDF have issue(s) with Typography options (Hide Icons & Underline Links)


Samrat-Saha-Sammy:
### Is there an existing issue for this?

- [X] Yes, I have searched the existing issues and none of them match my problem.

### Product Variant

Cloud (http://rxresu.me)

### Current Behavior

[_Note_] All Templates are affected. 

**Typography > Options > Hide Icons**

When the `Hide Icons` toggle option is enabled, the icons (eg. inside the header section @ icon, location icon etc. ) should not be visible or in other words should be hidden. 

**Inside Builder Preview**: Works as expected and looks fine (icons toggle visibility)
**Inside Downloaded PDF**: Not Working as properly as displayed in Builder Preview before download. When the `Hide Icons` toggle is disabled or Not hiding icons, it works. But when the toggle is enabled or hides the icons, the icon placeholders take a blank space, resulting in an unexpected User Interface Design. 

**Typography > Options > Underline Links**

When the `Underline Links` toggle option is enabled, the links (eg. inside the header section phone no., email etc. ) should be underlined. 

**Inside Builder Preview**: Works as expected and looks fine (links toggle underline)
**Inside Downloaded PDF**: Not Working as properly as displayed in Builder Preview before download. When the `Underline Links` toggle is disabled or Not underlining links, it works. But when the toggle is enabled or the user needs to underline the links, it is missing, resulting in unexpected behaviour. 

**Bug Demo** 

![Demo](https://github.com/AmruthPillai/Reactive-Resume/assets/4963559/b3d67096-948d-4a44-a778-d8e6117037df)

### Expected Behavior

**Typography > Options > Hide Icons**
**Inside Downloaded PDF**: When the toggle is enabled or hides the icons, the icon placeholders should not take a blank space. 

**Typography > Options > Underline Links**
**Inside Downloaded PDF**: When the toggle is enabled or the user needs to underline the links, then respected links should be displayed as underlined.

### Steps To Reproduce

1.  Open [rxresu.me](https://rxresu.me) in Browser
2.  Login with credentials
3.  Go to `Dashboard` Page
4.  Click on `Create a new resume`
5.  Click on the dropdown icon beside the `Create` button and click `Create Sample Resume` 
6.  Click on the newly created sample resume to open in editor/builder
7.  Navigate to `Typography Section` from the right sidebar
8.  Scroll down to the options section
9.  Toggle Options
[...]

PR #1751, opened 2024-02-01
fix(#1750): fixed downloaded PDF have issue(s) with Typography options

Samrat-Saha-Sammy:
The typography option toggle class were added in the DOM #root but when the server is printing the resume, a child section of the DOM #root is cloned and inserted for pdf processing which misses out the Typography option toggle class in the parent DOM, resulting in the bug, Thus, the same class were added to each cloned DOM node (a node for each page DOM with [data-page] attribute).


AmruthPillai:
Thank you for the fix! <3

Issue #1764, opened 2024-02-11
Link is overlaping in personal projects


bhumit070:
### Is there an existing issue for this?

- [X] Yes, I have searched the existing issues and none of them match my problem.

### Product Variant

Cloud (http://rxresu.me)

### Current Behavior

The link is overlapping like this in the personal projects section

![image](https://github.com/AmruthPillai/Reactive-Resume/assets/63059063/538af97b-93b6-4e6f-b4a4-cab2010cfbe5)

### Expected Behavior

It should not go beyond the sidebar

### What browsers are you seeing the problem on?

Chrome

### What template are you using?

Gengar

### Anything else?

Please assign this issue to me I will fix it and create a PR

PR #1768, opened 2024-02-13
fix(gengar template): link of personal project was overflowing outside.

bhumit070:
In Gengar template the link for personal project was overflowing out of sidebar so added a little CSS to fix the issue 
please see images attached below to get more clarity on issue and fix

This PR is for issue #1764 

Issue:
image

Fix:
image

### DIFF
--- a/apps/artboard/src/styles/main.css
+++ b/apps/artboard/src/styles/main.css
@@ -21,5 +21,5 @@
 }
 
 .wysiwyg {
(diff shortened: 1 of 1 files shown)


AmruthPillai:
Thank you for this fix, probably useful, but some folks might not want to break a word mid-way either. I think a better approach is to use break-words. Will experiment with it, but thank you for the PR, it's good to merge as-is! :)

PR #1779, opened 2024-02-18
fix(section list): add delete button

dhavalsavalia:
Here is a list view:

Delete Modal:

I am working on `section-dialog` to bring `Delete` button there too.


AmruthPillai:
The thing is, when a dialog is already open, clicking on delete would trigger another dialog to be open (on top of the older one, which seems counter productive). This is why I didn't add a delete button there in the first place.

I can explore other options to include the delete button inside the modal (maybe a one-click delete), but for now, will close this PR.

PR #1785, opened 2024-02-21
modify the name field to a minimum of 2 characters

tsp36:
There are a lot of 2-character names in China.

### DIFF
--- a/libs/dto/src/user/user.ts
+++ b/libs/dto/src/user/user.ts
@@ -15,7 +15,7 @@ export const usernameSchema = z
 
 export const userSchema = z.object({
   id: idSchema,
-  name: z.string().min(2).max(255),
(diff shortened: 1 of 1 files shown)


AmruthPillai:
I'll do you one better. I'll just change it to 1 :) Thank you for bringing this to my notice.

libs/dto/src/user/user.ts
```diff
@@ -15,7 +15,7 @@ export const usernameSchema = z
 
 export const userSchema = z.object({
   id: idSchema,
-  name: z.string().min(3).max(255),
+  name: z.string().min(2).max(255),
```
amruthpillai:
```suggestion
  name: z.string().min(1).max(255),
```

Issue #1754, opened 2024-02-03
Cannot download the resume I created


schwammbob:
Hi there - love this app, by the way.....I created my resume using one of the templates, but I cannot download it as a PDF.  I clicked on the button several times, looked on my desktop and download folders, and nothing.  Can you help?  I am using a MacBook Air and the Safari browser.

Thanks.

Stan

ayDavidGitHere:
I am facing the same issue. Clicking download pdf returns a notification "Request Failed with staus code 404"

ayDavidGitHere:
I went back to https://rxresu.me/dashboard/resumes and opened the resume, then I was able to print. 
Hope it works for you

Weav3r:
After hours of trying suggestions from other issues I got it working by modifying the compose file as follows:

1. Change the `PUBLIC_URL: http://localhost:3000` to point to the docker service `PUBLIC_URL: http://app:3000` 
2. Modify `STORAGE_URL: http://localhost:9000` by appending `default` to the url i.e.  `STORAGE_URL: http://localhost:9000/default` 

If you're using the `traefik.yml`  compose file the same steps applies. FYI, I'm running it locally and I haven't tested it on the cloud but I presume it should work. Also, I haven't really started extensively using the app so I don't know if anything else is broken but currently resume previews and pdf exports works. Hope this helps.

chunchunmaru0:
> After hours of trying suggestions from other issues I got it working by modifying the compose file as follows:
> 
> 1. Change the `PUBLIC_URL: http://localhost:3000` to point to the docker service `PUBLIC_URL: http://app:3000`
> 2. Modify `STORAGE_URL: http://localhost:9000` by appending `default` to the url i.e.  `STORAGE_URL: http://localhost:9000/default`
> 
> If you're using the `traefik.yml` compose file the same steps applies. FYI, I'm running it locally and I haven't tested it on the cloud but I presume it should work. Also, I haven't really started extensively using the app so I don't know if anything else is broken but currently resume previews and pdf exports works. Hope this helps.


This only works on local.
Tested on prod env, still throws the same issue. 
Tried with auth cookie and tested in local, works. I don't think it's an issue with compose in local.

amruthpillai:
Closing this issue as I'm doing some clean up to move to a newer version of the project. I apologize in advance for the inconvenience if you find the need to reopen this issue.

Please read #2499 for more information about the migration.

PR #1789, opened 2024-02-23
fix issue where PDF downloads fail on compose setup

bennycornelissen:
Implements the fix proposed in #1754 

This PR updates the`STORAGE_URL` for the `simple.yaml` Compose setup that has no reverse proxy, by adding the bucket name as a URL suffix. Without it, PDF downloads don't work.


AmruthPillai:
Good fix, it's very annoying that AWS/S3 buckets have a different naming structure and minio has it different. This really screwed me towards the end there when I was trying to get everything set up. Thank you!

PR #1799, opened 2024-03-06
fix(github_auth): fix GithubStrategy not searching for user using username

### LINKED ISSUE #1798: [Bug] Github authentication fails if user changes their primary email

### Product Variant

Cloud (http://rxresu.me)

### Current Behavior

I had earlier used Github to signup on Reactive Resume and was able to login normally. However I am unable to sign in since I changed my primary email address on Github, receiving the following error:
```json
{"message":"UserAlreadyExists","error":"Bad Request","statusCode":400}
```

### Expected Behavior

It should log into my account using Github without any issues.

### Steps To Reproduce

1. Signup on rxresu.me using Github
2. Change your Github email address
3. Log out of rxresu.me
4. Try to sign in back again using the same Github account
5. Error:
```json
{"message":"UserAlreadyExists","error":"Bad Request","statusCode":400}
```

Firefox, Chrome, Safari, Microsoft Edge

None

### Anything else?

The `GithubStrategy.validate` method only passes the user's email to `findOneByIdentifier`. Since it doesn't find my new email already registered, it tries to create a new user which fails since the username already exists, hence returning `UserAlreadyExists`.
```ts
    try {
      const user = await this.userService.findOneByIdentifier(email);

      if (!user) throw new UnauthorizedException();

      done(null, user);
    } catch (error) {
      try {
        user = await this.userService.create({
          email,
          picture,
          locale: "en-US",
          name: displayName,
          provider: "github",
          emailVerified: true, // auto-verify emails
          username: processUsername(username ?? email.split("@")[0]),
          secrets: { create: {} },
        });

        done(null, user);
      } catch (error) {
        throw new BadRequestException(ErrorMessage.UserAlreadyExists);
      }
    }
```

The `findOneByIdentifier` method actually supports both username and email, which should be used in case the user is not found by one of those options:
```ts
  async findOneByIdentifier(identifier: string) {
    const user = await (async (identifier: string) => {
      // First, find the user by email
      const user = await this.prisma.user.findUnique({
        where: { email: identifier },
        include: { secrets: true },
      });

      // If the user exists, return it
      if (user) return user;

      // Otherwise, find the user by username
      // If the user doesn't exist, throw an error
      return await this.prisma.user.findUniqueOrThrow({
        where: { username: identifier },
        include: { secrets: true },
      });
    })(identifier);

    if (!user.secrets) {
      throw new InternalServerErrorException(ErrorMessage.SecretsNotFound);
    }

    return user;
  }
```

mradigen:
This PR will close #1798

### DIFF
--- a/apps/server/src/auth/strategy/github.strategy.ts
+++ b/apps/server/src/auth/strategy/github.strategy.ts
@@ -34,7 +34,9 @@ export class GitHubStrategy extends PassportStrategy(Strategy, "github") {
     if (!email) throw new BadRequestException();
 
     try {
-      const user = await this.userService.findOneByIdentifier(email);
+      const user =
+        (await this.userService.findOneByIdentifier(email)) ||
+        (await this.userService.findOneByIdentifier(username));
 
       if (!user) throw new UnauthorizedException();
 


AmruthPillai:
Ah, good one. Thank you for the fix!

Issue #1800, opened 2024-03-06
[Feature] Allow search Simple Icons on Profiles


LitoMore:
### Is there an existing issue for this feature?

- [X] Yes, I have searched the existing issues and it doesn't exist.

### Feature Description

Currently, the Simple Icons input will send a request to `cdn.simpleicons.org` every time press the key. It's not a good practice sending too many 404 requests.

It would be good to create a searchable dropdown for showing all icons. Here is a similar thing: https://litomore.github.io/simple-icons-figma/. And it's source code is at https://github.com/LitoMore/simple-icons-figma

We could update the UI to use a [Popover](https://www.radix-ui.com/themes/docs/components/popover) on the icon and remove the input field.

Rash-Hit:
@AmruthPillai .. will it be fine .. if i implement debouncing for this issue ?? if yes then please assign this issue to me

AmruthPillai:
Honestly, for the sake of quicker reactivity, it's fine to hit a few 404s before reaching the correct URL. There's no harm in this.

LitoMore:
There are bunch of requests on this UI. And it's consuming the Vercel resource of https://github.com/LitoMore/simple-icon-cdn rapidly. You may reach the rate limit in a short time.

Better to use a debounce for web requests to keep your stability.

AmruthPillai:
@LitoMore Ah, I'm sorry. I had no idea that you were one of the maintainers of Simple Icons. Has this app caused so much of a hassle for you guys at Simple Icons? Just surprised by the reach is all. So sorry about that. Thank you so much for all your work with maintaining the icon library. Hope it's not been too much trouble.

If it's okay to share, I'd be curious to know how much of the traffic on Vercel for the CDN was from rxresu.me?

@Rash-Hit Feel free to come up with a solution for debouncing the requests made to the Simple Icons CDN. (300-400 ms should do, I think). Will assign the ticket to you :)

LitoMore:
> If it's okay to share, I'd be curious to know how much of the traffic on Vercel for the CDN was from rxresu.me?

I can't see this due to we're using Vercel's free plan.

BTW, I can help to create a dropdown for your when I have time. So that users can preview icons in a list.

Rash-Hit:
@AmruthPillai migrations command not working , can you please help ??

LitoMore:
@Rash-Hit Please check if your step 4 is correct.

[...]

PR #1809, opened 2024-03-11
debounce in profile icon

Rash-Hit:
fixes : #1800 

https://github.com/AmruthPillai/Reactive-Resume/assets/73944177/3c4497a2-0a36-49a5-ab42-c444aeebad4f

### DIFF
--- a/apps/client/src/pages/builder/sidebars/left/dialogs/profiles.tsx
+++ b/apps/client/src/pages/builder/sidebars/left/dialogs/profiles.tsx
@@ -32,18 +32,7 @@ export const ProfilesDialog = () => {
 
   const [iconSrc, setIconSrc] = useDebounceValue("", 400)
 
-  const handleIconChange = useCallback((event: React.ChangeEvent<HTMLInputElement>, field: ControllerRenderProps<{
-    id: string;
-    visible: boolean;
-    network: string;
-    username: string;
-    icon: string;
-    url: {
-      label: string;
-      href: string;
-    };
-  }, "icon">) => {
-    field.onChange(event);
+  const handleIconChange = useCallback((event: React.ChangeEvent<HTMLInputElement>) => {
     if (event.target.value === "") {
       setIconSrc("");
     } else {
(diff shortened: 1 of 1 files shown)


Rash-Hit:
@AmruthPillai lmk , if any changes required ..

apps/client/src/pages/builder/sidebars/left/dialogs/profiles.tsx
```diff
@@ -28,6 +30,29 @@ export const ProfilesDialog = () => {
     resolver: zodResolver(formSchema),
   });
 
+  const [iconSrc, setIconSrc] = useDebounceValue("", 400)
+
+  const handleIconChange = useCallback((event: React.ChangeEvent<HTMLInputElement>, field: ControllerRenderProps<{
+    id: string;
```
amruthpillai: You could re-use the FormValues type here instead of defining the type all over again.

apps/client/src/pages/builder/sidebars/left/dialogs/profiles.tsx
```diff
@@ -83,14 +108,14 @@ export const ProfilesDialog = () => {
...
+                    {iconSrc && (
                       <AvatarImage
                         className="p-1.5"
-                        src={`https://cdn.simpleicons.org/${field.value}`}
+                        src={iconSrc}
                       />
                     )}
                   </Avatar>
-                  <Input {...field} id="iconSlug" placeholder="linkedin" />
+                  <Input {...field} id="iconSlug" placeholder="linkedin" onChange={(e) => handleIconChange(e, field)} />
```
amruthpillai: Personally, if the only reason to pass `field` to the `handleIconChange` function is to call `field.onChange(event)`, you could just create an anonymous function inline to call it separately. So it would look something like this:

```suggestion
                  <Input {...field} id="iconSlug" placeholder="linkedin" onChange={(event) => {
                    field.onChange(event);
                    handleIconChange(event);
                  }} />
```

This way, you don't need to pass the field (and it's complex types) to the function.

AmruthPillai (changes requested):
Great initial code, great that you made use of the existing `useDebounceValue` hook from the library, that's good thinking for sure.

Just a couple of suggestions here.

Rash-Hit:
@AmruthPillai sir can you please review it ..

AmruthPillai:
It's perfect, will get it to production soon enough :)

PR #1820, opened 2024-03-20
fix(auth): E-Mail capitalization issue

### LINKED ISSUE #1740: [Bug] E-Mail Login is dependent on capitalization

### Product Variant

Cloud (http://rxresu.me)

### Current Behavior

When you register with your e-mail address and there is a capital letter in the mail address, there will be a server error when you login with the same e-mail with a different capitalization

### Expected Behavior

E-Mail login and registration should be completely indepentend from capitalization in the e-mail address.

### Steps To Reproduce

1. Visit rxresu.me
2. Create an account with a mail that has at least one capital letter
3. Try logging into that account in the start page again but change the capitalization in the e-mail address

Chrome

None

### Anything else?

_No response_

DollyXSoul:
Fixes   #1740

This issue affects email in three places where user needs to enter the email - during registration ,login and forgot-password page.
To fix that we convert the email to lowercase before processing or sending to db.

-> During the user registration , the email entered by user is converted to lowercase and then stored in db.

->Similarly , in forgot-password , so again as we do a check in db if given email exists or not, so again make given user email to lowercase.

-> And in login user can enter email or username(already in lowercase) so if we convert user-input to lowercase it won't affect login using username.

### DIFF
--- a/apps/client/src/pages/auth/forgot-password/page.tsx
+++ b/apps/client/src/pages/auth/forgot-password/page.tsx
@@ -35,6 +35,8 @@ export const ForgotPasswordPage = () => {
   });
 
   const onSubmit = async (data: FormValues) => {
+     //email convert to lowecase
+     data.email = data.email.toLowerCase();
     await forgotPassword(data);
 
     setSubmitted(true);
--- a/apps/client/src/pages/auth/login/page.tsx
+++ b/apps/client/src/pages/auth/login/page.tsx
@@ -41,6 +41,8 @@ export const LoginPage = () => {
   });
 
   const onSubmit = async (data: FormValues) => {
+    //email/username convert to lowecase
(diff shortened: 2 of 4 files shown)


AmruthPillai:
@DollyXSoul Wouldn't there be some need to case-insensitivize the backend auth checks as well?

DollyXSoul:
@AmruthPillai Yes you are right I forgot about that. Should I just convert email to lowercase on backend as well  , or you have a better suggestion to do this ??

AmruthPillai:
I tried running this locally, and ran into some issues when trying to login with an existing account that had capitalization already stored in the database. For now, I'm going to close the PR as I feel this is one place in the app where it might need some regression tests.

But thank you very much for your efforts. It is very much appreciated.

Issue #1831, opened 2024-03-27
[Bug] Two-Factor Authentication (2FA) Setup Verification Zod Message is incorrect


lplawlor:
### Is there an existing issue for this?

- [X] Yes, I have searched the existing issues and none of them match my problem.

### Product Variant

Cloud (http://rxresu.me)

### Current Behavior

Upon setting up a TOTP device, you must "Enter the 6-digit code from your authenticator app to verify that 2FA has been setup correctly." However, if you submit an incorrect code, the message displayed in the [`FormMessage`](https://github.com/AmruthPillai/Reactive-Resume/blob/6e6914fe6ba4e9aea157d1cc062950e5fcfaec82/apps/client/src/pages/dashboard/settings/_dialogs/two-factor.tsx#L203) component is a strange 6-character string.

As an example, I created a dummy account and put in just the first 5 digits of the 2FA code (I've since deleted the account so this screenshot isn't at risk of revealing anything sensitive.)
![image](https://github.com/AmruthPillai/Reactive-Resume/assets/60618151/42770d2c-e8e9-4c89-a1ab-b8ac73250efe)

### Expected Behavior

The error message should read "Code must be exactly 6 digits long.", as defined in the [Zod schema](https://github.com/AmruthPillai/Reactive-Resume/blob/6e6914fe6ba4e9aea157d1cc062950e5fcfaec82/apps/client/src/pages/dashboard/settings/_dialogs/two-factor.tsx#L50).

### Steps To Reproduce

1. Go to User settings while logged-in
2. Disable 2FA (if already enabled).
3. Click "Enable 2FA"
4. Setup your TOTP device.
5. Enter an incorrect code into the "Code" field on the dialogue which reads "Enter the 6-digit code from your authenticator app to verify that 2FA has been setup correctly."

### What browsers are you seeing the problem on?

Firefox

### What template are you using?

None

### Anything else?

[...]

PR #1843, opened 2024-04-02
Fix(tfa): error message for tfa code length

abizek:
Related to #1831


AmruthPillai:
Thank you for finding a fix for this, very much appreciated! :)

PR #1844, opened 2024-04-02
Fix(builder): right sidebar scroll to page section

abizek:
Fixes an issue where clicking on the page icon in the right sidebar doesn't scroll to the page section, unlike the other icons.


AmruthPillai:
Ah, good catch. Thank you!

Issue #1847, opened 2024-04-02
[Bug] Sidebar is not stretching to the full height


JKamsker:
### Is there an existing issue for this?

- [X] Yes, I have searched the existing issues and none of them match my problem.

### Product Variant

Cloud (http://rxresu.me)

### Current Behavior

![grafik](https://github.com/AmruthPillai/Reactive-Resume/assets/11245306/2d6440ea-abda-48f2-8b2f-a46b85f6db2e)
The sidebar does not stretch

### Expected Behavior

The sidebar should stretch from top to bottom

### Steps To Reproduce

1. Select Chikorita
2. Enter some text, but not enough to fill the page
3. The sidebar is not the full length

### What browsers are you seeing the problem on?

Firefox

### What template are you using?

Chikorita

prithvi5481:
Can you please assign this issue to me as it seems it needs a small change in styling. Can give min width of 100% and this would work probably.

JKamsker:
@prithvi5481 I mean, feel free to fork and pr a solution

AashishSinghal:
I have raised fix PR (here)[https://github.com/AmruthPillai/Reactive-Resume/pull/1891]

PR #1848, opened 2024-04-02
Fix sidebar height

JKamsker:
Related: https://github.com/AmruthPillai/Reactive-Resume/issues/1847
Probably not the right way of fixing this but it looks like it's fixed, even with more or less items in it. 
If anyone got a better idea how to fix it, please tell me 😅


JKamsker:
Oh thanks for the tip, will do!

Issue #1683, opened 2023-12-20
[Bug] Link for documentation for local deployment?


justpankajdotcom:
### Is there an existing issue for this?

- [X] Yes, I have searched the existing issues and none of them match my problem.

### Product Variant

Cloud (http://rxresu.me)

### Current Behavior

I tried to search for documentation for local self-hosted deployment in both open and closed issues and found couple of links, however, they all redirect to 404 - Page not found. Is the address changed? I can see comments from author in other issues mentioning the below links as latest as april-may 2022. Is the address changed after that?

I cannot find any relevant links for deployment here --> [https://docs.rxresu.me/](https://docs.rxresu.me/)

[https://docs.rxresu.me/deployment/](https://docs.rxresu.me/deployment/) (This link mentioned by author in one of the issue for deployment doc doesn't work anymore)

### Expected Behavior

The link should direct to deployment page

### Steps To Reproduce

Go to [https://docs.rxresu.me/deployment/](https://docs.rxresu.me/deployment/)

### What browsers are you seeing the problem on?

Chrome

### What template are you using?

None

justpankajdotcom:
> I also couldn't find any recent/up-to-date docs, so I just thought for self-hosted option we'd probably have to figure it out ourselves by checking the [`tools/compose`](https://github.com/AmruthPillai/Reactive-Resume/tree/main/tools/compose) directory for examples, in combination with the [`.env.example`](https://github.com/AmruthPillai/Reactive-Resume/blob/main/.env.example). Still, some environment variables are not mentioned in the `.env.example` (like the one mentioned in #1640), so it's a bit of a puzzle to get it setup properly 🧩
> 
> Docs would definately help future users 👌

Yes, exactly. I tried with different setups but i cant get it working somehow. If you are also trying and haven't been able to yet. Shall we connect somewhere outside, like discord if it is okay for you? Maybe with collaboration it will become easier.

ToshY:
Hey @justpankajdotcom 👋 

I've finally got it fully working yesterday evening locally 🙂 Sure, I'm open to discuss it on Discord (see my github profile for username) and share my `docker-compose.yml` and `.env.example` with you before posting it here.

justpankajdotcom:
Hi @ToshY ,

Appreciate it. But I can't seem to search you there in discord with your username. Can you please check your privacy settings once?

ToshY:
[...]

PR #1849, opened 2024-04-02
Fix storageurl in dev

JKamsker:
Probably fixes https://github.com/AmruthPillai/Reactive-Resume/issues/1683

Problem for me was, that the "download pdf" button redirected me to the right url, but the wrong port after changing the storage port in the ``.env`` file

### DIFF
--- a/apps/server/src/printer/printer.service.ts
+++ b/apps/server/src/printer/printer.service.ts
@@ -192,6 +192,13 @@ export class PrinterService {
       await page.close();
       browser.disconnect();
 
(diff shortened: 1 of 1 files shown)


AmruthPillai:
The issue is, if you're not running in `Development` mode, this code block never gets accessed, and NODE_ENV is set to development only when running the app using `pnpm dev`.

The port is not always included in the storage URL. For example, if the storage service is running on a seperate domain/IP, on either port 80 or 443, then you wouldn't need a port. This is why I initially recommended to add the port in the STORAGE_URL env. This works no matter what environment you are running in.

PR #1856, opened 2024-04-09
feat(server): Support for SMTP-over-TLS

WangZhiYao:
As per the code restriction specified in the configuration schema,  the `SMTP_URL` was required to start with `smtp://`.

https://github.com/AmruthPillai/Reactive-Resume/blob/fe77b14807873efb12af535fd119b2cbe17d87e9/apps/server/src/config/schema.ts#L29

However, my email service provider utilizes SSL on port 465. Despite attempting various configurations based on the nodemailer documentation available [here](https://nodemailer.com/smtp/), I encountered difficulties in sending emails through the SMTP service using the following settings:

```
smtp://username:password@smtp.example.com:465
smtp://username:password@smtp.example.com:465/?secure=true
smtp://username:password@smtp.example.com:587/?requireTLS=true
```

Upon further investigation, I found that nodemailer supports the `smtps://` protocol. Consequently, I have updated the validation in zod to permit the `SMTP_URL` configuration to start with `smtps://`. This adjustment now enables me to utilize the following configuration for sending emails via SMTP of my email service provider:

```
smtps://username:password@smtp.example.com:465
```

So I made this pull request, hoping to help people who have the same problem as me.

### DIFF
--- a/apps/server/src/config/schema.ts
+++ b/apps/server/src/config/schema.ts
@@ -26,7 +26,7 @@ export const configSchema = z.object({
 
   // Mail Server
   MAIL_FROM: z.string().includes("@").optional().default("noreply@localhost"),
-  SMTP_URL: z.string().url().startsWith("smtp://").optional(),
+  SMTP_URL: z.string().url().refine(url => url.startsWith("smtp://") || url.startsWith("smtps://")).optional(),
 
   // Storage
   STORAGE_ENDPOINT: z.string(),


AmruthPillai:
Seems like a valid PR, although the validation check could just be simplified to `.startsWith("smtp")` because it would include both smtp and smtps.

WangZhiYao:
> Seems like a valid PR, although the validation check could just be simplified to `.startsWith("smtp")` because it would include both smtp and smtps.

If change this to `startsWith("smtp")` instead of fully qualifying it as the protocol schema, then `smtp+tls://` will also pass the validation check but nodemailler does not seem to support such a protocol, this may cause some other confusions, so I don't recommend for this.

AmruthPillai:
Agreed, looks good to me! :)

PR #1885, opened 2024-05-07
Fix: generate PDF name from title instead of id

### LINKED ISSUE #836: [FEATURE] PDF takes the same name as the resume from dashboard
**Is your feature request related to a problem? Please describe.**
When downloading the PDF, I noticed it has random characters as the filename.

**Describe the solution you'd like**
When rendering a new PDF, take the name of the resume that we assigned in the dashboard.

**Describe alternatives you've considered**
Maybe append datetime after the filename in case people want to download multiple version and don't want it to overwrite the original PDF?

nickconway:
Fixes #836 

This changes the generated PDF name from the ID to the title (e.g. Resume V34.pdf).

### DIFF
--- a/apps/server/src/printer/printer.service.ts
+++ b/apps/server/src/printer/printer.service.ts
@@ -187,7 +187,7 @@ export class PrinterService {
         resume.userId,
         "resumes",
         buffer,
-        resume.id,
+        resume.title,
       );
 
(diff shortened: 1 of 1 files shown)


AmruthPillai:
Sounds good to me, name is probably better than ID.

PR #1978, opened 2024-06-27
fix(edit profile picture): fix the page crash happening when clicking on circular border radius

### LINKED ISSUE #1953: [Bug] Profile Picture aspect ratio adjustment crash

### Product Variant

Cloud (https://rxresu.me)

### Current Behavior

I was experimenting with the profile picture aspect ratio when I selected circular shape. And when I tried reselect after clicking the aspect ratio box it crashed.

https://github.com/AmruthPillai/Reactive-Resume/assets/24765995/e0d36f8e-d872-4d2e-9c9b-8ce1a3f93f48

image

### Expected Behavior

Should not crash, in best case should reset the aspect ratio settings

### Steps To Reproduce

1. Select Circular shape for profile photo
2. Click in the Aspect ratio box
3. Click on the circular shape, 
4. it should crash

Chrome

None

### Anything else?

_No response_

googleknight:
Fixes  #1953 

### Issue
In profile picture edit section, when a user selects a circle border radius, clicks on the numeric input for the border radius, and then clicks outside, the `onBorderRadiusChange` function is triggered and receives an empty string (`""`) as the `value`.

### Fix
So updated the code to check for empty string as being done in `onAspectRatioChange` function.

### DIFF
--- a/apps/client/src/pages/builder/sidebars/left/sections/picture/options.tsx
+++ b/apps/client/src/pages/builder/sidebars/left/sections/picture/options.tsx
@@ -61,8 +61,9 @@ export const PictureOptions = () => {
     return borderRadiusToStringMap[radius];
   }, [picture.borderRadius]);
 
-  const onBorderRadiusChange = (value: BorderRadius) => {
-    setValue("basics.picture.borderRadius", stringToBorderRadiusMap[value]);
+  const onBorderRadiusChange = (value: string) => {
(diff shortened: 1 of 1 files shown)


googleknight:
@AmruthPillai Could you please review the fix ?
Thanks!

AmruthPillai:
Good solution, thank you for the fix!

Issue #1998, opened 2024-07-22
[Feature] URL support for Custom Fields


StijnGroenen:
### Is there an existing issue for this feature?

- [X] Yes, I have searched the existing issues and it doesn't exist.

### Feature Description

Like #1743, I would like my profiles to be in the header section of my CV.
With #1780, it is already possible to add custom icons to the header section by using custom fields.
The only thing that is missing is URL support.
It would be great if URLs in a custom field would automatically become links.

PR #1999, opened 2024-07-22
feat(templates): added url support for custom fields

StijnGroenen:
Issue #1998
Added URL support for custom fields.
If the value of a custom field is a URL, the custom field will render as a link with the provided name as the title. If there is no name, it will show a clickable URL.

Before:
image
After:
image


AmruthPillai (approved):

AmruthPillai:
Looks great, thank you!

lmssiehdev:
is this out in production yet? doesn't work for me.

raymondyangdev:
> is this out in production yet? doesn't work for me.

Not yet. I was experiencing the same issue but realised that this feature was merged in July, but the last release was 4.1.7 in June. If you want to use this feature, you'll need to run it locally for now.

barrelltitor:
I actually do not want this behavior, is this not toggle-able? The name and URL can't stay separate?

PR #2008, opened 2024-08-04
fix(docs): Add missing themes to bug-report

busches:
A few themes were missing when I filled out a bug report.


AmruthPillai:
Thank you for fixing that up :)

PR #2009, opened 2024-08-04
fix(docs): Update CONTRIBUTING.md for Node 20

busches:
package.json and GitHub Actions both run with Node 20. Originally changed in 748b509b1db5725f122c463c5ad02926b80cccf4


AmruthPillai:
Good find, thank you for your PR!

PR #2017, opened 2024-08-14
feat(toggle-pan): Add toggle to switch between "Scroll to Pan" and "Scroll to Zoom"

infinia-yzl:
### Motivation:
To improve UX with trackpads.

### Explanation:
The current trackpad behaviour does not feel natural. Usually when scrolling on a trackpad using 2 fingers, the page or content would Pan; whereas Zooming is achieved by pinching 2 fingers.

### Considerations:
On a mouse, it is arguably more user-friendly to Scroll to Zoom. Therefore, I have added a toggle button to switch between behaviours.

I opted for the default behaviour to be Scroll to Pan because there are already various methods to control the zoom levels, and I believe that panning actions are more common than zooming. For context, the default Double-click behaviour is set to Zoom In; and there are the Zoom In & Out buttons on the toolbar to control the zoom levels. Open to discussion / change.

### Changes:
- Add a new button to the Toolbar to handle switching between "Scroll to Pan" and "Scroll to Zoom"
- Modify wheel behaviour to pan by default for a more natural behaviour with trackpads (https://github.com/BetterTyped/react-zoom-pan-pinch/pull/447)
- Bump react-zoom-pan-pinch version to `3.6.1` from `3.4.4`

### Screenshot:
New button
image

### DIFF
--- a/apps/artboard/src/pages/builder.tsx
+++ b/apps/artboard/src/pages/builder.tsx
@@ -1,7 +1,7 @@
 import { SectionKey } from "@reactive-resume/schema";
 import { pageSizeMap, Template } from "@reactive-resume/utils";
 import { AnimatePresence, motion } from "framer-motion";
-import { useEffect, useMemo, useRef } from "react";
+import { useEffect, useMemo, useRef, useState } from "react";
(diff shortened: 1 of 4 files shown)


AmruthPillai:
I like the idea, will make sure to goes on rxresu.me very soon! :)

PR #2020, opened 2024-08-20
feat: script to update PNPM dependencies and start up docker-based install

theschles:
I wrote this to make it easier to get the dependencies in place and start up a docker-based install.  What do you think?


AmruthPillai:
Not required at the moment, it seems to me. A better option would be a detailed guide in the documentation on how to set up the project with docker locally.

Issue #2034, opened 2024-09-06
[Bug] Resume cards are clipping on hover in the dashboard page grid view


noreb001:
### Is there an existing issue for this?

- [X] Yes, I have searched the existing issues and none of them match my problem.

### Product Variant

Cloud (https://rxresu.me)

### Current Behavior

When I hover or press on a card then some of that card is clipped off when the animation is run. On mobile this happens to each card. On desktop, this only clips the first or last card in a row. 

Desktop
![image](https://github.com/user-attachments/assets/16ca8861-34a6-4d01-9cd5-d3294c8c64b2)

Mobile: 
![image](https://github.com/user-attachments/assets/f3c3f6a5-4ed8-4831-876a-2e98c06aaa34)

### Expected Behavior

Cards should not clip and the whole card should be visible when the animation is run 

### Steps To Reproduce

1. Log in and go to dashboard page 
2. Click or hover over a card that's next to a border

### What browsers are you seeing the problem on?

Firefox, Chrome, Safari, Microsoft Edge

### What template are you using?

None

noreb001:
[PR raised for this](https://github.com/AmruthPillai/Reactive-Resume/pull/2035)

AmruthPillai:
fixed in https://github.com/AmruthPillai/Reactive-Resume/pull/2035

PR #2035, opened 2024-09-06
bugfix: Fixed clipping issue with cards on dashboard page

noreb001:
[Issue](https://github.com/AmruthPillai/Reactive-Resume/issues/2034)

Fixed Card clipping issue on Dashboard page in grid view. Where the sides of the card where being clipped once the hover animation is triggered. 

**BEFORE**

desktop
image

Mobile
image

**AFTER**
Desktop
image

Mobile
image

**Explanation:**
By default ScrollAreaPrimitive.Root sets the class name to 'overflow-hidden'. Then passing a classname with 'overflow-visible' to the scroll-area UI component is not enough. "ScrollAreaPrimitiv.Viewport" applies some extra default styles if the props type='scroll'. I have added a boolean prop type that allows us to disable this and allow bot overflowX and overflowY to be visible in the scenarios where we need it. 

Let me know if you need more information.

### DIFF
--- a/apps/client/src/pages/dashboard/resumes/page.tsx
+++ b/apps/client/src/pages/dashboard/resumes/page.tsx
@@ -49,7 +49,10 @@ export const ResumesPage = () => {
           </TabsList>
         </div>
 
-        <ScrollArea className="h-[calc(100vh-140px)] lg:h-[calc(100vh-88px)]">
+        <ScrollArea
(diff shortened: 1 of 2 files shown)


AmruthPillai:
Really good solution, thank you for the work put into this. I'll make sure it hits rxresu.me very soon! :)

PR #2042, opened 2024-09-10
Support for getting webSocketDebuggerUrl from /json/version

LichtHong:
For newer versions of Chrome, the webSocketDebuggerUrl is usually generated at startup as a dynamic UI, and the current webSocketDebuggerUrl can be obtained via GET /json/version.


AmruthPillai:
Not sure why this is required, since the version of Chrome is fixed (and we only access the interface through browserless). But thank you for your contribution!

Issue #2047, opened 2024-09-13
[Bug] Awards Summary Text Color is different than Title/Other fields


ChrisDeardeuff:
### Is there an existing issue for this?

- [X] Yes, I have searched the existing issues and none of them match my problem.

### Product Variant

Self-Hosted

### Current Behavior

When adding text to the summary field in the Awards section, it is the same color as the main font but not the title/other fields.  When this is on a colored background (as with the Chikorita template, it makes it difficult to see and the color can't be changed without changing all of the main sections fonts. 
![image](https://github.com/user-attachments/assets/6f1e1492-0a9f-49ee-958f-98a334d2c31e)

### Expected Behavior

Summary text is the same color as other fields in the Awards section

### Steps To Reproduce

Choose a template with a colored area (like Chikorita),
Add an award, fill in the title and a summary

### What browsers are you seeing the problem on?

Chrome

### What template are you using?

Chikorita

ChrisDeardeuff:
This looks like it will be fixed by PR: #2048 , any estimate on when this will be merged? Thanks :smiley:

PR #2048, opened 2024-09-13
fix(templates): fix Summary text in the Awards section to be the same…

mohab-kamle:
fix issue #2047
[Bug] Awards Summary Text Color is different than Title/Other fields 
after fixing image
Screenshot 2024-09-14 025309


DemaPy (commented):
class removed

AmruthPillai:
Looks good but went with another solution, but thank you for the contribution! :)

PR #2091, opened 2024-11-05
fix(templates): Add right border after website in Rhyhorn

### LINKED ISSUE #2090: [Bug] No right border after website on Rhyhorn if custom fields are used

### Product Variant

Cloud (https://rxresu.me)

### Current Behavior

Currently, if there is a website in the Profile/Basics section and a custom field(s) is also added, there is no right border after the website to separate the website and the custom field. 

This only seems to affect the Rhyhorn template.

### Expected Behavior

There should be a right border separator to separate the website and the first custom field.

### Steps To Reproduce

1. Select Rhyhorn as template
2. Add website to Profile section
3. Add a custom field

Chrome

Rhyhorn

### Anything else?

_No response_

raymondyangdev:
Closes #2090 
Added right border to website if custom fields are used

Before:
Screenshot 2024-11-05 at 5 10 21 PM
After:
Screenshot 2024-11-05 at 5 11 14 PM

### DIFF
--- a/apps/artboard/src/templates/rhyhorn.tsx
+++ b/apps/artboard/src/templates/rhyhorn.tsx
@@ -124,7 +124,7 @@ const Link = ({ url, icon, iconOnRight, label, className }: LinkProps) => {
   if (!isUrl(url.href)) return null;
 
   return (
-    <div className="flex items-center gap-x-1.5">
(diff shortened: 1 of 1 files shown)


AmruthPillai:
Good catch, thank you for the fix!

PR #2095, opened 2024-11-07
fix(printer): refine "localhost" matching in printer.service.ts 

jbilinski:
Bug: unable to print if url contains the word "localhost" in any location 
resulting in the "about:blank", no output path
example: `http://nonlocalhoster.com` would be changed to `http://nonhost.docker.internaler.com`

proposed: 
change localhost matching for local dev environments from a url string match to more specific regex that matches host property.

 - Low Priority
 - Edge Case
 - Service resilience

### DIFF
--- a/apps/server/src/printer/printer.service.ts
+++ b/apps/server/src/printer/printer.service.ts
@@ -101,17 +101,17 @@ export class PrinterService {
 
       let url = publicUrl;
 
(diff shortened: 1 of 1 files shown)


jbilinski:
this was an issue for me, but i think unlikely for many others

AmruthPillai:
Looks good, thank you for your contribution!

Issue #2101, opened 2024-11-13
[Feature] Make main take full width if sidebar is empty


0x346e3730:
### Is there an existing issue for this feature?

- [X] Yes, I have searched the existing issues and it doesn't exist.

### Feature Description

The idea would be to automatically make the main part take full-width of the page if the sidebar part is empty on the page.

Without this feature :
![image](https://github.com/user-attachments/assets/b54f396d-c172-4382-9894-23bd2274dd54)

With this feature :
![image](https://github.com/user-attachments/assets/ba58d484-fe4e-4456-9a55-b331d9c4be59)

I have a PR incoming that implements this behavior.

PR #2103, opened 2024-11-14
feat(templates): make main full-width if sidebar is empty

0x346e3730:
Fixes https://github.com/AmruthPillai/Reactive-Resume/issues/2101

Adjust templates so that the main part takes full-width of the page if there is no sidebar on the page, to avoid having blank spaces


AmruthPillai:
Love the idea, merging it right away. Thank you for the contribution!

PR #2134, opened 2025-01-05
fix(docs): Update privacy-policy.md

DemaPy:
Replace [insert contact email] with hello@amruthpillai.com


AmruthPillai:
Ah, thank you for the catch! :)

PR #2137, opened 2025-01-11
fix(builder): skills and languages form error from level data type

GrinZero:
# Desc
When I adjust the "level" in the skills pop-up, clicking save will pop up a form to check for errors: 
image

The actual reason is that the onChange and onValueChange of the Slider component take effect simultaneously


GrinZero:
😅

PR #2181, opened 2025-01-26
Make /v1 endpoint optional in order to support Ollama models

vvok12:
[Earlier commit](https://github.com/AmruthPillai/Reactive-Resume/commit/18cf814779474063f7edba1958a91f9d5f8c233e) removes the ability to specify /v1 which breaks Ollama integration. 

This PR fixes regression by making /v1 ending optional 

## Summary by CodeRabbit

- **Bug Fixes**
	- Updated OpenAI settings URL validation to support optional `/v1` endpoint segment
	- Improved base URL validation for more flexible configuration


AmruthPillai:
@vvok12 Can we update the PR to just check if the URL starts with https or http? So the url path or other routes just shouldn't matter. This might be useful when v1 changes to v2 someday and we can make use of the app even then.

PR #2205, opened 2025-02-03
fix(linkedin): use case-insensitive slug check to prevent missing image

ImDarkShadow:
**Summary:**
Fixes an issue where varying capitalizations of "linkedin" (e.g., "LinkedIn") resulted in an empty image by making the slug check case-insensitive.

**Changes Made:**

- Converted the slug to lowercase before comparison to handle all capitalization variations.

### DIFF
--- a/apps/artboard/src/components/brand-icon.tsx
+++ b/apps/artboard/src/components/brand-icon.tsx
@@ -5,7 +5,7 @@ type BrandIconProps = {
 };
 
 export const BrandIcon = forwardRef<HTMLImageElement, BrandIconProps>(({ slug }, ref) => {
-  if (slug === "linkedin") {
+  if (slug.toLowerCase() === "linkedin") {
     return (
       <img
         ref={ref}


skatkoria:
@ImDarkShadow @ilatypov @gzsombor 
https://github.com/AmruthPillai/Reactive-Resume/issues/2210

I have raised defect since week but no response over it. I am actively applying for the job and require resume export from your website https://rxresu.me/ 
Can you please take priority for my defect ? or suggest any workaround.

ImDarkShadow:
Hi @skatkoria,

I’m not an active contributor to this project, and my knowledge of it is somewhat limited. However, I noticed your comment and attempted to reproduce the issue, but was unable to do so. If you could provide the relevant JSON, I’d be happy to take another look and see if there’s anything I can help with.

Please note that this is the Pull Request section, not the Issues section. For future reference, issues should be discussed in the appropriate section.

Additionally, I would recommend using a different template for PDF export, as this project is maintained by a sole developer, and responses may take some time.

Thank you for your understanding!

PR #2219, opened 2025-02-21
fix(templates): don't show section when all items are hidden

mtdvlpr:
## Summary by CodeRabbit

- **Refactor**
	- Refined the display logic for content sections so that only sections containing visible items are rendered. This update improves the consistency of the interface by ensuring that sections without any visible content remain hidden.


DemaPy:
@mtdvlpr isn't would be better to move it into ```utils``` folder? 
This is clearly duplicated logic.

Issue #2245, opened 2025-03-21
[Feature] Display language name and description on the same line


encarbassotnopot:
### Is there an existing issue for this feature?

- [x] Yes, I have searched the existing issues and it doesn't exist.

### Feature Description

I hide the level bar in the languages section, and I'd like to have each language's name and description in one line. This will reduce the amount of empty space on the sidebar and make it more coherent with the other sections.
This could be a toggle, as it isn't incompatible with having the level bar displayed.

This is how languages and descriptions are currently displayed on RxResume:
![Image](https://github.com/user-attachments/assets/72752edb-3c4b-4303-8152-bc30a0d69452)

And this is what I'd like to achieve:
![Image](https://github.com/user-attachments/assets/f16cbce4-f0c8-41d2-9771-8d301727dc95)

yphanvansong:
You can manage this with custom CSS.


```
#languages .font-bold{
    display:inline;
}
#languages .font-bold+div{
    display:inline;
}
```

You have to add separators to avoid the two elements to touch.

PR #2246, opened 2025-03-21
feat(lang-section): option to show language name and description on a single line

encarbassotnopot:
taking a first stab at closing #2245


yphanvansong:
You can manage this with custom CSS.


```
#languages .font-bold{
    display:inline;
}
#languages .font-bold+div{
    display:inline;
}
```

You have to add spaces to avoid the two elements to touch.

Issue #2371, opened 2025-08-25
[Bug] Summary section sits only on the main side in the first page


SidKarthik1437:
### Is there an existing issue for this?

- [x] Yes, I have searched the existing issues and none of them match my problem.

### Product Variant

Cloud (https://rxresu.me)

### Current Behavior

Summary section sits only on the main side in the first page and wont shift to the sidebar section. Meanwhile it works properly in other pages

### Expected Behavior

Should be able to move the section to the sidebar even on the first page

### What browsers are you seeing the problem on?

Chrome, Firefox

### What template are you using?

Gengar

PR #2412, opened 2025-10-12
fix(template): fix gengar template summary section not rendering in sidebar

alay-dev:
The Summary section remains in the main area and cannot be moved to the sidebar, unlike on other pages. (See issue: #2371)

✅ What changed / Fix
Modified the layout logic so the Summary section is eligible to be moved into the sidebar even on page one.

🔗 References
Fixes issue #2371

## Summary by CodeRabbit

* **New Features**
  * Added a dedicated Summary section accessible via sidebar navigation and app routing.

* **Refactor**
  * Ensures Summary is rendered consistently through the app’s section mapping by removing the prior first-page special-case.

### DIFF
--- a/apps/artboard/src/templates/gengar.tsx
+++ b/apps/artboard/src/templates/gengar.tsx
@@ -531,6 +531,7 @@ const mapSectionToComponent = (section: SectionKey) => {
     case "education": {
       return <Education />;
     }
+    
     case "summary": {
       return <Summary />;
     }


apps/artboard/src/templates/gengar.tsx
```diff
@@ -531,6 +531,10 @@ const mapSectionToComponent = (section: SectionKey) => {
     case "education": {
       return <Education />;
     }
+    case "summary": {
```
amruthpillai, line 535:
```suggestion
    
    case "summary": {
```

Issue #2431, opened 2025-11-03
[Feature] Using `puppeteer-core` instead of `puppeteer`


Ofsen:
### Is there an existing issue for this feature?

- [x] Yes, I have searched the existing issues and it doesn't exist.

### Feature Description

Hello!

I stumbled upon this project very recently and I find it very cool!

I noticed that the `puppeteer` package is used and all the docker compose configurations have the `browserless/chromium` in them.

The `puppeteer` package uses the `puppeteer-core` library and also downloads a chrome for testing version ( checkout [https://pptr.dev/guides/installation](https://pptr.dev/guides/installation) ). So one improvement that can be done is either use the downloaded one or just use `puppeteer-core` and keep everything else as is.

PR #2432, opened 2025-11-03
chore(puppeteer): using puppeteer-core instead of puppeteer

Ofsen:
Removing `puppeteer` and using `puppeteer-core` instead.
Read more here: [https://github.com/AmruthPillai/Reactive-Resume/issues/2431](https://github.com/AmruthPillai/Reactive-Resume/issues/2431)

## Summary by CodeRabbit

* **Chores**
  * Switched the browser-automation dependency to a lighter core package and bumped its version.
  * Updated workspace build configuration to reflect the dependency change for consistent builds.


amruthpillai:
@Ofsen Thank you for the contribution. Looks like a good change, have you verified whether the PrinterService still works after replacing it with puppeteer-core?

Ofsen:
@AmruthPillai I didn't fully test it, developement env is weird, had issues with the `waitForSelector` and `host.docker.internal` not working ( maybe because am on ubuntu ) but I started the app using the `simple.yml`, then created a sample resume and exported it to pdf. It gave me a pdf but not separated by pages and the cover avatar image didn't fully load: [grieving-sporting-giraffe.pdf](https://github.com/user-attachments/files/23336720/grieving-sporting-giraffe.pdf)
Maybe the pages are not separated because I added the custom section of lorem's ?

amruthpillai:
That works, I think there's a different issue regarding loading the image. But this change is a welcome one, thank you so much.

Issue #2482, opened 2025-12-20
[Feature] Permalink to share/download resume json data


y8a5y:
### Is there an existing issue for this feature?

- [x] Yes, I have searched the existing issues and it doesn't exist.

### Feature Description

Exactly like the "Sharing" option, but to expose the json data of the resume. The URL would be something like `<username>/<slug>/json`.

Can be useful for things like automatic build/updates of a webpage with the up-to-date resume fields, or to easily parse the data with external tools.

y8a5y:
Closed issue #2088 seems similar but failed to deliver the feature I'm describing here, manual json export was good enough for OP.

y8a5y:
I'm currently trying to implement it, but there is several ways to do it, each having pros and cons. I'm thinking of three ways to do it:

1. add another toggle **inside** the "Sharing" section (assuming feature will stay similar to the normal share, without evolving much in the future).
2. add the option as another section, **under** "Sharing" (assuming the feature can evolve into something more complicated later).
3. add the option under the json export inside the "Export" section (not my favourite since the the layout risk to be too messy).

For now, I did not take statistics into much consideration. In my opinion it should not count towards view count, and another statistic may be added later if need be.

y8a5y:
note: feature was validated and merge successfully, will wait for the change to be available on prod before closing this issue.

amruthpillai:
Closing this issue as I'm doing some clean up to move to a newer version of the project. I apologize in advance for the inconvenience if you find the need to reopen this issue.

Please read #2499 for more information about the migration.

PR #2483, opened 2025-12-21
[Feature] Permalink to share/download resume json data

y8a5y:
## Why?

As described in #2482, this feature aims to make the raw json data of a resume easily available to external tools.

Some example use cases:

- automatically building/updating a webpage with the most up-to-date info (what I needed this feature for),
- talk more easily with an LLM, if you're into that.

## How?

The most easily way to do it was to use the URL of the corresponding API request, and to make the link available directly in the Sharing section as quality of life improvement.

### Problem

What took the most effort was to not break statistics, the incrementation of resume view was done directly in the route I needed to use.

## Final solution

I added a new endpoint, simply adding `/stealthily` at the end as differentiation. This new endpoint returns exactly the same data, without triggering a view count.

### What needed to be done

Since the increment was done inside the `findOneByUsernameSlug` method, I moved the increment logic outside, as a new method in `resume.service.ts`. The original route then calls this method when an increment is needed, the final logics stayed exactly the same. The new stealth route simple do not call the increment method.

After that, all remaining was to add a new card inside the Sharing section of a resume. The new URL, named "Raw Json URL", is formed inside the component directly, like the normal share URL.

I added a small label indicating that the new URL is for advanced users and that it will not count towards statistics.

> Note: the json returned by the new URL is not the same as a json export from the Export section: the new URL returns the unformated API data, whereas the export strips some info (e.g. id, timestamps, ...).

## Additional change

The visible caret made the URLs feels like they were editable, even tho the input is read only. Not much, but I made the caret transparent in both the URL boxes, making them less confusing.

## End result

new card also depend on the public sharing toggle:
image

new card when public sharing is enable:
image

(note that the first URL input is selected in this picture, yet the caret isn't visible, less confusing imho)

toast when link is copied:
image

## Summary by CodeRabbit

* **New Features**
  * Added a Raw JSON export URL and dedicated copy button in the sharing panel for public resumes.
  * Added a stealthy public access option that returns resumes with private notes hidden.

* **Bug Fixes / Behavior Changes**
  * View counting refined to consistently record views for unauthenticated visitors.

* **Refactor**
  * Centralized private-notes redaction and view-counting logic for public resume responses.

* **Documentation**
  * CONTRIBUTING updated with recommended lint, test, and build checks before commits.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -82,4 +82,14 @@ git commit -m "fix(homepage): fix typo on homepage in the faq section"
 
 It helps to be as descriptive as possible in commit messages so that users can be aware of the changes made by you.
 
+Then, you can run the basic checks, as to make sure the CI has the most chance of being successful:
+
+```
+pnpm run lint
+
+pnpm run test
+
+pnpm run build
+```
+
 Finally, create a pull request to merge the changes on your forked repository to the original repository hosted on AmruthPillai/Reactive-Resume. I can take a look at the changes you've made when I have the time and have it merged onto the app.


amruthpillai:
Nice work, thank you very much for your contribution! Looks like a solid feature implemention.

Could you make sure that the CI pipeline passes so I can merge?

y8a5y:
Thanks for the approbation!

I'll try to tackle the CI today.

y8a5y:
@AmruthPillai everything should be ok, I've run the basic CI steps (`lint`, `test` and `build`) locally.

I've also took this as an opportunity to add those steps to the CONTRIBUTING.md, to make sure future PRs have the most chance of not wasting CI runs.

Issue #2486, opened 2025-12-24
Page Break with A4 size


PrasadBattena:
### Is there an existing issue for this?

- [x] Yes, I have searched the existing issues and none of them match my problem.

### Product Variant

Cloud (https://rxresu.me)

### Current Behavior

When we download the document in PDF format, it is not downloading in A4 size. Additionally, the file is generated as a single page instead of two pages.

Please find the attached copy below for your reference.

[prasad (1).pdf](https://github.com/user-attachments/files/24328470/prasad.1.pdf)

### Expected Behavior

IT SHOULD BE IN 2 PAGES IF THE CONTENT GOES BEYOND ONE PAGE

### What template are you using?

None

MdTanwer:
@PrasadBattena Hi, I’ve gone through this issue and would love to work on it. Kindly assign it to me.

amruthpillai:
Closing this issue as I'm doing some clean up to move to a newer version of the project. I apologize in advance for the inconvenience if you find the need to reopen this issue.

Please read #2499 for more information about the migration.

PR #2489, opened 2025-12-31
fix(PrinterService): enhance PDF generation by adding page height calculations and handling multiple pages

szympol:
@AmruthPillai PR for respecting multi PDF pages print.

- **Printer service**: Honor the new metadata.page.options.paginate flag. When paginate=true, wait for fonts/images, set viewport to the chosen A4/Letter width/height, clip the long artboard into page-sized slices, and merge them into a multi-page PDF. When paginate=false, keep the legacy behavior by rendering one full-height PDF page (no slicing).
- Add **paginate page option** (default true) to metadata schema/defaults and expose it in builder sidebar + toolbar (Files icon) so users can switch between multi-page PDFs and the legacy single long page.
- **Improve builder page guides:** repeat dashed break lines at every page height and align sizing with the page size constants.

Note: 
- The pagination toggle is optional—happy to remove it if it doesn’t fit UX expectations. 
- In multi-page mode, extremely long blocks can sometimes split across the page boundary so a single line gets cut in half (top on page one, bottom on page two); adding a little manual spacing (e.g., in a job description) can prevent that.

Ref: [Issue: 2486](https://github.com/AmruthPillai/Reactive-Resume/issues/2486)

image

[primary-vocational-stork.pdf](https://github.com/user-attachments/files/24395875/primary-vocational-stork.pdf)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,7 @@ node_modules
 *.launch
 .settings/
 *.sublime-workspace
+.history
 
 # IDE - VSCode
 .vs/*
--- a/apps/artboard/src/components/page.tsx
+++ b/apps/artboard/src/components/page.tsx
@@ -1,5 +1,5 @@
 import { useTheme } from "@reactive-resume/hooks";
-import { cn, pageSizeMap } from "@reactive-resume/utils";
+import { cn, MM_TO_PX, pageSizeMap } from "@reactive-resume/utils";
 
 import { useArtboardStore } from "../store/artboard";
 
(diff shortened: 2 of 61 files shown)


remotehive-dev (approved):

mels0n:
Really looking forward to this addition. The resume being 'unprintable' has come up before.

> * In multi-page mode, extremely long blocks can sometimes split across the page boundary so a single line gets cut in half (top on page one, bottom on page two); adding a little manual spacing (e.g., in a job description) can prevent that.

If you do something like this in the generateResume function when building for A4 or Letter you won't need to manually interact with this spacing (this is a page size to printable size issue.) The code will scale it to 90% center of page, which effectively introduces 'margins' to avoid printer hardware issues like this.

        await page.addStyleTag({
          content: `
            [data-page] > div {
              transform: scale(0.9);
              transform-origin: center center;
              width: 100%;
            }
          `,
        });

amruthpillai:
Closing this PR as I'm doing some clean up to move to a newer version of the project. I apologize in advance for the inconvenience if you find the need to reopen the PR.

Please read #2499 for more information about the migration.

PR #2556, opened 2026-01-22
chore: clean up Goternberg references

VinhLe1410:
### What's changed?
- Just a small fix in the `compose.yml` file and `vite-env.d.ts`. They're still pointing to the Goternberg container, but that's being replaced with Browserless.


amruthpillai:
Good catch, thank you for this!

Issue #2537, opened 2026-01-20
[Bug] Username is invalid


kalwadi:
### Is there an existing issue for this?

- [x] Yes, I have searched the existing issues and none of them match my problem.

### Product Variant

Self-Hosted

### Describe the bug you're experiencing

not sure if this is undocumented feature or a bug, but username doesn't support having `-`

my first user is usually `username-admin` which is not supported, `username_admin` worked and I successfully created the user

### What template are you using?

None

amruthpillai:
Oh, interesting. I didn't know about this limitation myself. It's probably coming from Better Auth. I can look into whether this would be an easy thing to remove, or if it has side effects. Thank you for bringing this to my attention.

divyanshu-patil:
hello sir,
i have opened a PR which fixes this issue,
it was problem with better-auth as by default it doesn't allows hyphens in username

the issue is fixed in #2558 

thank you

PR #2558, opened 2026-01-23
fix: username not allowing hyphens

divyanshu-patil:
closes https://github.com/amruthpillai/reactive-resume/issues/2537

better auth default validation only allows `alphanumeric characters and underscores` not hyphens

we need to run custom validator after that to allow hyphens too

image


amruthpillai:
Perfect, thank you for the PR, really helps in taking some of the workload off.

PR #2583, opened 2026-01-25
Improved the Right to Left Direction implementation for relevant languages.

obreo:
Hi,
I have added a number of changes to the Text Editor, UI, and the Page Templates that focus on solving the right-to-left direction content. The changes are described as follows:

1. Added right-to-left support for Bullet Points and Numbered Points in the Text Editor when Right Align is clicked, which was not implemented before.

Before:
image

After:
image

2. Added Toggle Button at the Layout Section that flips the templates to the right direction. In addition to the specified CSS fixes per template that changed the border lines to be dynamic for left and right instead of being fixed for the left direction only.

The Template direction change is implemented by adding `style=" direction: rtl;"` to the template dev element inside the page dev element. This is controlled by the toggle and defaults to the left direction (OFF)

When Toggle is OFF (Default):
image

When Toggle is ON:
image

3. Minor changes to CSS syntax for some templates to be RTL aware instead of being fixed to left only with styles like:
margin/padding-inline-start/end,  border-inline-start/end

instead of:
border/margin/padding-left

So now elements such as the border in the image attached and circle are dynamic in either direction and not fixed to the left only:

image

These changes have been tested locally for all the templates and work perfectly.

Kindly merge this feature to help right to left langauge users from Hebrew, Arabic and Persian write resumes in their languages.


amruthpillai:
Very interesting changes, and really nice implementation of it all. I was just wondering, whether it is necessary to add a switch/toggle to enable/disable RTL?

The alternative approach, if I could suggest, is to use the `metadata.page.locale` from the resume, which determines the language the resume is written is, and we can use the `isRTL` function from `@/utils/locale`:

```
const RTL_LANGUAGES = new Set([
	"ar", // Arabic
	"ckb", // Kurdish (Sorani)
	"dv", // Dhivehi
	"fa", // Persian
	"he", // Hebrew
	"ps", // Pashto
	"sd", // Sindhi
	"ug", // Uyghur
	"ur", // Urdu
	"yi", // Yiddish
]);

export function isRTL(locale: string): boolean {
	const language = locale.split("-")[0].toLowerCase();
	return RTL_LANGUAGES.has(language);
}
```

This way, we can auto-detect the RTL support without having to switch it back on/off. I use this already in the rich-input component and the html `dir` attribute.

I don't really speak any RTL languages, so I very much appreciate the help and support with this PR. Thank you very much for the proactive effort!

obreo:
Thanks for the suggestion! Actually this is a cleaner approach! I have modified the implementation as suggested.

So currently when the user changes the Language to RTL based language in the Page section, the template will switch right to left as same as the toggle/switch feature. 

The Toggle has been removed including the code related to it and the langauge translation of the Toggle lable has been cleared.

The pull request have been updated!

amruthpillai:
Awesome, I'll get this merged ASAP and have it shipped with the v5.0.3 release. I also added another commit which basically replaced all usages of margin-left, margin-right, padding-left, padding-right etc. to their start/end equivalents in CSS. This too should hopefully help in overall RTL accessibility.

Thank you for your work on the rich input editor especially, wouldn't have thought of doing that, but it's quite important.

obreo:
Thanks! @amruthpillai

Issue #2564, opened 2026-01-23
[Feature] The arrangement of the two columns in the left-hand menu for organising the sections


SvenT22:
### Is there an existing issue for this feature?

- [x] Yes, I have searched the existing issues and it doesn't exist.

### Feature Description

The arrangement of the two columns in the left-hand menu for organising the sections should be identical to that of the CV templates. There, for example, the skills section is usually on the left and the work experience on the right. In the left-hand menu, this is reversed. This can be confusing. There are currently two template exceptions where the arrangement is identical to that in the page menu. If necessary, the page menu can be made dynamic here so that the columns correspond to the template arrangement. 

<img width="1646" height="693" alt="Image" src="https://github.com/user-attachments/assets/274330ae-8d24-4e81-bc4a-9e691c99b8a8" />

amruthpillai:
Interesting concept. I understand your point of view on this, and it would probably be possible by identifying which templates are RTL/LTR (or `hasLeftSidebar`) and toggling this view. I just wonder if it would be confusing for the user to see the layout section jump around as well. But I'll think about it.

Thank you for the amazing feedback loop by the way, very much appreciate all the issues and inputs ❤️

ffischbach:
Hey @amruthpillai , I was also bothered by this, so I went ahead and implemented a fix. Looking forward to your thoughts on the PR!

Also, I noticed that we currently don't merge sidebar sections into the main section when switching templates (e.g., from a template with a sidebar to a full-width one). If we implement that, we could also hide the full-width toggle whenever a template without a sidebar is active to clean up the UI.

I'd be happy to take a look at both points after this PR, if you'd like.

https://github.com/amruthpillai/reactive-resume/pull/2603

PR #2603, opened 2026-01-26
feat(builder): arrange sidebar builder depending on section type in template #2564

ffischbach:
The current layout of the section management in the left-hand menu often mirrors the template structure incorrectly. In most templates, the Skills section is on the left and Work Experience is on the right, but the menu currently shows them in reverse order. This inconsistency can be confusing for users.

This PR aligns the column arrangement in the menu with the actual visual representation in the CV templates.

Changes:
- UI Alignment: Swapped/Adjusted the column order in the section management menu to match the template layouts.
- Consistency: Ensured that the mental model of "left side in menu = left side on CV" holds true across the application.

Related Issues
Closes [#2564](https://github.com/amruthpillai/reactive-resume/issues/2564)

### DIFF
--- a/src/dialogs/resume/template/data.ts
+++ b/src/dialogs/resume/template/data.ts
@@ -2,18 +2,12 @@ import type { MessageDescriptor } from "@lingui/core";
 import { msg } from "@lingui/core/macro";
 import type { Template } from "@/schema/templates";
 
-export enum SidebarType {
-	NO_SIDEBAR,
-	LEFT,
-	RIGHT,
-}
-
 export type TemplateMetadata = {
 	name: string;
 	description: MessageDescriptor;
 	imageUrl: string;
 	tags: string[];
-	sideBarType: SidebarType;
+	sidebarPosition: "left" | "right" | "none";
 };
 
 export const templates = {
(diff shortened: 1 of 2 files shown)


amruthpillai:
Looks great, just made a couple of changes for easier readability.

Also, about the full width toggle idea. Marking as full width actually does more than just hiding the sidebar section, it also moves all items from the sidebar to the `main` swimlane. This is essential for the resume template to have all sections. Right now, if we simply hid the section, there would be no way modify the positioning of these items.

Also, if we were to push items from sidebar to main when someone changes their template, it would lose the original history of how their layout was, so if they were to simply click on a template to try it out, it's harder to go back to the original because they would have to set up their layout all over again. 

Anyway, thank you so much for your PR, I'll make sure to get it on https://rxresu.me very soon!

ffischbach:
Removing the 'Full Width' toggle definitely has more implications than I initially realized.

I’ve added a TODO to investigate whether there’s a cleaner way to handle this in the future—perhaps by implementing a layout history or state management that remembers where items belong. However, I agree that the effort might outweigh the benefits right now. It would require some serious thought to ensure we don't end up confusing the users.

Also, thank you for the refactoring! The code looks significantly better now. :)

PR #2652, opened 2026-01-30
Update PostgreSQL volume path in compose.yml

lazy-media:
added `data` path to postgres; was causing this to create a throw away database upon restarting app, therefor causing this act as a new install every time app was restarted. Now data is persisted (at least for me)

### DIFF
--- a/compose.yml
+++ b/compose.yml
@@ -9,7 +9,7 @@ services:
       POSTGRES_USER: postgres
       POSTGRES_PASSWORD: postgres
     volumes:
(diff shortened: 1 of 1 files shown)


amruthpillai:
The current implementation is actually correct, and the recommended method for Postgres 18. More information about it here: https://www.reddit.com/r/selfhosted/comments/1nsjxm2/if_youre_moving_to_docker_postgres_18_you_should/

The image you were running locally was probably <18, in which case, you do need the /data suffix.

PR #2672, opened 2026-02-01
fix(layout): handle text overflow in sidebar/footer elements

abdulalim110:
What does this PR do? This PR replaces truncate / text-nowrap with break-words whitespace-normal to improve text wrapping in the sidebar and footer sections.

Why is this needed? The current layout often hides information or breaks the design when using languages with specific characteristics:

Finnish: Contains very long compound words without spaces, which previously got cut off.

Indonesian: Often has long addresses or descriptions that need to wrap to the next line naturally.

Japanese: Uses characters without standard spacing, causing layout breaks in narrow containers.

How was it tested? I verified the changes using dummy data in the languages mentioned above. The text now wraps correctly to the next line and breaks long words only when necessary to fit the container.

Screenshoots:

Finnish

before
before-finnish

after
after-finnish

Japanese

before
before-japanese

after
after-japanese

Indonesian

before
before-indonesian

after
after-indonesian

### DIFF
--- a/src/routes/dashboard/-components/sidebar.tsx
+++ b/src/routes/dashboard/-components/sidebar.tsx
@@ -179,7 +179,7 @@ export function DashboardSidebar() {
 							animate={{ y: 0, height: "auto", opacity: 1 }}
 							exit={{ y: 50, height: 0, opacity: 0 }}
 						>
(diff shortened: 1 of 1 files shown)


amruthpillai:
Good catch, thank you for this fix!

Issue #2677, opened 2026-02-02
[Feature] Chore: Update `schema/resume/sample.ts` with cover letter and optional fields


a-earthperson:
### Is there an existing issue for this feature?

- [x] Yes, I have searched the existing issues and it doesn't exist.

### Feature Description

Since a cover letter was added in #2645, I am just updating the sample resume with it, along with exercising some optional fields in the `ResumeData` schema that were already available, but no example for them existed.

Also helps me with #2604.

a-earthperson:
I'll close this issue now.

PR #2675, opened 2026-02-02
Feature/resume drafting api

a-earthperson:
Related to issue https://github.com/amruthpillai/reactive-resume/issues/2646

This is a large PR; I tried hard to keep it low‑risk. The intent is *separation*, not behavioral change: no workflow changes, no frontend changes, no schema migrations, no DB rewrites. `Resume.View` is meant to remain a drop‑in replacement for the existing `ResumeData` shape.

## Overview
We are splitting the current monolith (`ResumeData`) into two orthogonal concerns:
- **`Resume.Draft`** — content only (data fields).
- **`Resume.Styles`** — presentation only (visual configuration).
- **`Resume.View`** — a zip/merge of `(Draft, Styles)` that preserves the existing `ResumeData` contract.

This keeps data and style independent, while still composing the exact payload shape the app and templates already consume.

### 1) Schema split + view composition
- Introduced `Resume.Draft` and `Resume.Styles` schemas/types.
- Added `Resume.View` schema/types that **compose** draft + styles.
- Added factories to **zip** / **unzip** / **default** these structures.
- Added tests to validate **parity** between `Resume.View` and legacy `ResumeData`.

### 2) Draft API (separate surface area)
I have implemented a completely separate set of routes under `/draft` that support batch edit operations at `/draft/{id}/ops`. REST API clients can make complex edits in one atomic call. When finished, they can call the `/draft/{id}/finalize` (also atomic and idempotent) endpoint, which *promotes* that `Resume.Draft` to a `Resume.View`, by applying default `Resume.Styles`. The newly promoted `Resume.Draft` will now be visible under `/resume/{id}`.
- Added `/draft` endpoints for Draft lifecycle operations, independent of `/resume`.
- Goal: REST API clients can iteratively build drafts without touching resume styles until finalization.
- The frontend doesn't even need to know about this endpoint.
- `/draft` API auth scope is same as `/resume`.

> **Note:** Some draft endpoints are *planned* but not final yet.

### 3) `/resume/{id}` behavior remains unchanged
- Once a draft is finalized into a view, the existing `/resume/{id}` flow remains intact.
- The frontend editor can keep issuing `PUT` updates (content + styles) with no changes required.

## Example Workflow
```mermaid
sequenceDiagram
autonumber
participant Client as REST Client
participant API as API (/draft, /resume)
participant Draft as Resume.Draft
participant View as Resume.View
participant FE as Frontend Editor

Client->>API: POST /draft/create (DraftData)
API->>Draft: create DraftData record
[...]

PR #2678, opened 2026-02-02
update sample with cover letter and other existing fields

a-earthperson:
Update schema/resume/sample.ts with cover letter and optional fields. Related to #2677. Helps me with reasoning about other changes in #2675. 

I tested that nothing breaks by creating a new sample resume and toggling/viewing the changed/added fields in the sample.


amruthpillai:
Good idea, thanks for the PR, it should help a lot in discovery of the feature.

Issue #2687, opened 2026-02-06
[Feature] `PUT` endpoint on `/resume/{id}` should return updated document


a-earthperson:
### Is there an existing issue for this feature?

- [x] Yes, I have searched the existing issues and it doesn't exist.

### Feature Description

Currently, successfully making a `PUT` request on the `/resume/{id}` endpoint results in an empty response object. This is fine, but it requires making another `GET` request to know the state of the backend. It's a valid design pattern when there is only one API client (the frontend), but this isn't true all the time.

a-earthperson:
PR #2688 implements this.

a-earthperson:
closing this since #2688 was merged.

PR #2688, opened 2026-02-06
 Return updated resume object from `PUT /resume/{id}` instead of `void`

a-earthperson:
# Return updated resume on PUT /resume/{id}

## Summary
- Return the updated resume object from `PUT /resume/{id}` instead of `void`
- Reuse a shared response schema between `GET /resume/{id}` and `PUT /resume/{id}`
- Ensure clients can consume a canonical, persisted resume snapshot (including `hasPassword`)

## Problem / Motivation
The current `PUT /resume/{id}` handler returns `void`. That forces clients to either:
- perform an additional fetch for the updated record, or
- rely on their local optimistic state, which can diverge from the persisted data.

For example, the server computes `hasPassword` and the updated data may be normalized by persistence. Returning
the updated object enables immediate cache synchronization and avoids stale UI state after an update.

## Solution
1. Define a shared response schema (`resumeByIdOutputSchema`) for the full resume payload.
2. Change the `PUT /resume/{id}` output to use the shared schema.
3. After `resumeService.update(...)` completes, fetch and return the updated resume via
   `resumeService.getById(...)`.

## API Behavior (After Change)
**Endpoint:** `PUT /resume/{id}`  
**Response:** Resume object

Expected response shape:
- `id`
- `name`
- `slug`
- `tags`
- `data`
- `isPublic`
- `isLocked`
- `hasPassword`

Error cases remain unchanged:
- `RESUME_LOCKED` when attempting to update a locked resume
- `RESUME_SLUG_ALREADY_EXISTS` on conflicting slug per user

## Impact / Risk
- Adds a single, consistent read after each update to return the persisted record.
- No behavior changes to validation or error handling; only response shape changes.

### DIFF
--- a/src/integrations/orpc/services/resume.ts
+++ b/src/integrations/orpc/services/resume.ts
@@ -296,7 +296,16 @@ export const resumeService = {
 						eq(schema.resume.userId, input.userId),
 					),
 				)
-				.returning();
+				.returning({
+					id: schema.resume.id,
+					name: schema.resume.name,
+					slug: schema.resume.slug,
+					tags: schema.resume.tags,
+					data: schema.resume.data,
+					isPublic: schema.resume.isPublic,
+					isLocked: schema.resume.isLocked,
+					hasPassword: sql<boolean>`${schema.resume.password} IS NOT NULL`,
+				});
 
 			return resume;
 		} catch (error) {


a-earthperson:
Addresses issue #2687

amruthpillai:
Looks good, just made some slight changes to make the code quality a bit more stricter, going forward.

Also, instead of calling `getById` again, we can make use of drizzle's `.returning()` to return the newly updated record: https://github.com/amruthpillai/reactive-resume/pull/2688/changes#diff-fdde30d3ef8b47f466c16138dbbc2d81ea404005d927d9c37fe14085d68ae4e9R289

Thank you very much for the PR and all your help!

a-earthperson:
very nice! love the refactor 🤩

PR #2695, opened 2026-02-08
feat: add separate icon visibility toggles for headers and sections

sandipansingh:
## Summary
Previously, the resume settings provided only a single global option to hide all icons across the entire resume.

This PR introduces more specific controls to:

- Hide icons in headers
- Hide icons in sections

## Changes
- Added separate visibility toggles
- Updated UI settings panel
- Preserved existing functionality

## Motivation
Provides better customization and cleaner resume exports.


amruthpillai:
Seems like a large change to the schema that I'm not willing to implement at the moment before further testing. I would prefer adding an example in the documentation on how to hide icons per section through Custom CSS instead.

PR #2714, opened 2026-02-12
fix(auth): correct autocomplete semantics for login/register forms

GoldenSection0618:
This PR fixes browser autofill mis-mapping between auth forms (notably in Edge), where username could be filled into the email field.

## Changes

1. Updated login identifier field autocomplete to `section-login username`
2. Updated login password field autocomplete to `section-login current-password`
3. Updated register form autocomplete tokens to:
    - `section-register name`
    - `section-register username`
    - `section-register email`
    - `section-register new-password`

## Motivation

- The login identifier accepts either email or username, so using `email` autocomplete was misleading for password managers
- Adding `section-*` scopes prevents cross-form autofill leakage between sign-in and sign-up pages


amruthpillai:
It's a good change, always happy to see better accessibility. Thank you for the PR!

PR #2718, opened 2026-02-13
fix(auth): avoid dropping session_token Set-Cookie on sign-in

### LINKED ISSUE #2715: [Bug] Credentials(Username/Email + Password) sign-in returns 200 but no session cookie is set (stays on /auth/login)

### Product Variant

Self-Hosted

### Summary
Self-hosted v5 (main): **signup works**, but **username+password** and **email+password** sign-in do not log in. UI briefly refreshes/re-renders and remains on `http://localhost:3000/auth/login` with **no error shown**.

### Expected behavior
Successful sign-in sets an authenticated session and redirects to the app.

### Actual behavior
- Click **Sign In** → page refresh/re-render → still on `/auth/login`
- No toast/error
- API calls return **200**
- Session remains unauthenticated

### Steps to reproduce
1. Deploy self-hosted v5 main.
2. Register a new account successfully.
3. Go to `/auth/login`.
4. Sign in with valid **username+password** (also repro with **email+password**).
5. Observe: no redirect; still logged out.

None

#### POST /api/auth/sign-in/username
- Status: **200**
- Response header contains only:
  - `Set-Cookie: better-auth.dont_remember=; Max-Age=0; Path=/; HttpOnly; SameSite=Lax`
- **Missing** `Set-Cookie: better-auth.session_token=...`

#### GET /api/auth/get-session (immediately after)
- Status: **200**
- Same `dont_remember` cookie header; **no session cookie set**
- Session stays unauthenticated

GoldenSection0618:
Fixes #2715

### Problem
Self-hosted credentials sign-in can return `200 OK` but fails to establish a browser session:
- `better-auth.session_token` is not set
- `GET /api/auth/get-session` stays unauthenticated
- UI re-renders and remains on `/auth/login` (no visible error)

### Root Cause
When the sign-in response contains multiple `Set-Cookie` headers, the TanStack Start cookie bridge (`tanstackStartCookies()`) can end up forwarding only one `Set-Cookie`, dropping the `better-auth.session_token` entry under certain request states (e.g. when a stale session cookie already exists). Auth succeeds, but the session cookie never reaches the browser.

### Fix
- Remove `tanstackStartCookies()` from the Better Auth plugin list so the `session_token` cookie is not dropped during credentials sign-in (current flow is client-driven and does not depend on this bridge).

> Note: current credentials sign-in flow is client-driven and does not rely on TanStack Start server-side cookie bridging.

### Verification
1. Sign up a new user
2. Ensure a stale `better-auth.session_token` exists (or keep cookies from previous attempts)
3. Sign in via username/email + password
4. Confirm response includes `Set-Cookie: better-auth.session_token=...`
5. Confirm `GET /api/auth/get-session` returns a valid session and the app redirects out of `/auth/login`

### Notes
If server-side sign-in via `auth.api.signIn*` is introduced in the future and depends on TanStack Start cookie bridging, we may need a safer multi-`Set-Cookie` merge/append implementation instead of removing the plugin.

### Scope note (optional changes)
This PR includes two small, **optional** additions alongside the core fix, mainly to reduce confusion during self-hosted/local debugging. The **primary fix** is removing `tanstackStartCookies()` to prevent `session_token` `Set-Cookie` loss.

- **`trustedOrigins` (localhost ⇄ 127.0.0.1):** People often mix these loopback hosts in local/self-hosted setups. Treating them as trusted siblings avoids origin-mismatch cases that can look like “login bounces back to /auth/login”. Not required for the cookie-loss fix.
- **Login submit hardening + fallback toast:** When sign-in returns `200` but no session is established, the UI currently looks like a silent refresh. Minimal error handling/toast improves diagnosability without changing the auth flow; 2FA redirect behavior is preserved.

If you prefer minimal diffs, I can split the optional parts into a follow-up PR.


src/integrations/auth/config.ts
```diff
@@ -20,6 +19,27 @@ function isCustomOAuthProviderEnabled() {
...
+	};
+
+	// Allow localhost and 127.0.0.1 to be treated as trusted siblings in local development.
+	if (appUrl.hostname === "localhost") {
+		trustedOrigins.add(originWith("127.0.0.1"));
+	}
+
+	if (appUrl.hostname === "127.0.0.1") {
+		trustedOrigins.add(originWith("localhost"));
+	}
```
DemaPy:
```suggestion
    const LOCAL_ORIGINS = ["localhost", "127.0.0.1"];
	if (LOCAL_ORIGINS.includes(appUrl.hostname)) {
		trustedOrigins.add(originWith(appUrl.hostname));
	}
```

src/integrations/auth/config.ts
```diff
@@ -20,6 +19,27 @@ function isCustomOAuthProviderEnabled() {
 	return Boolean(env.OAUTH_CLIENT_ID) && Boolean(env.OAUTH_CLIENT_SECRET) && (hasDiscovery || hasManual);
 }
 
+function getTrustedOrigins() {
+	const trustedOrigins = new Set([env.APP_URL]);
+	const appUrl = new URL(env.APP_URL);
+	const originWith = (hostname: string) => {
```
DemaPy: Name of this funcion is not obvious.

What about changeHostname?

src/routes/auth/login.tsx
```diff
@@ -24,7 +23,7 @@ export const Route = createFileRoute("/auth/login")({
 
 const formSchema = z.object({
 	identifier: z.string().trim().toLowerCase(),
-	password: z.string().trim().min(6).max(64),
+	password: z.string().min(6).max(64),
```
DemaPy: why you have deleted trim function

src/integrations/auth/config.ts
GoldenSection0618: Good call, the old name was pretty vague.
I'll rename it to `withHostname` so it’s clearer what the helper does.

src/integrations/auth/config.ts
GoldenSection0618: Thanks for the suggestion — using a `LOCAL_ORIGINS` array is a nice cleanup. I’ll keep the localhost ↔ 127.0.0.1 sibling-trust behavior, but refactor it to the array/loop approach (skip self, add the sibling).

src/routes/auth/login.tsx
GoldenSection0618: Thank you for calling this out.
Sorry for mixing this into the same PR — I should’ve called it out more clearly.
The core fix for #2715 is the `tanstackStartCookies()` change.  
Removing `.trim()` on `password` was a small hardening, not the root-cause fix.

My goal was to avoid mutating the password on the client. `trim()` can change the actual credential (e.g. trailing spaces), and it also makes login behave differently from signup/reset. That said, I agree this is behavior-changing and likely out of scope for this bugfix PR.

If you prefer, I can revert the `.trim()` change from this PR. I’m happy to do that in a follow-up push.

amruthpillai:
Great PR, thank you so much for your contribution, and @DemaPy thank you so much for the assist in the review, really good comments that made this so much easier :) Will have this merged ASAP.

PR #2743, opened 2026-02-23
Resolve PDF download issue in offline mode

Nyndow:
## Description

This PR fixes an issue where PDF download fails in offline mode when running the app via Docker.

## Problem

When running the application offline (e.g., using Docker without internet access), 
the `setTimeout` delay of 500ms before triggering `downloadWithAnchor` was too short.

As a result: 
The download always returned error because of the timeout

## Solution: 
 Increased the timeout duration to allow sufficient time for the PDF to be prepared

## How to Reproduce

1. Run the app using Docker
3. Attempt to download a PDF
4. Observe that download may fail with the previous 500ms delay

## How to Test

1. Run the app in Docker (offline mode)
2. Download a PDF
3. Confirm that the file downloads successfully

### DIFF
--- a/src/utils/file.ts
+++ b/src/utils/file.ts
@@ -30,7 +30,7 @@ export function downloadWithAnchor(blob: Blob, filename: string) {
 	a.click();
 	document.body.removeChild(a);
 
-	setTimeout(() => URL.revokeObjectURL(url), 500);
+	setTimeout(() => URL.revokeObjectURL(url), 5000);
 }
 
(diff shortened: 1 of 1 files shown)


amruthpillai:
Not sure why increasing the timeout to 5 seconds specifically is helpful, but I'll welcome this change as I don't expect users to export a PDF multiple times within a span of 5 seconds, at least I hope they don't. Thank you for the PR!

PR #2770, opened 2026-03-03
fix(printer): make PDF margins match resume background color

### LINKED ISSUE #2741: [Bug] Margins don't match background colour.

### Product Variant

Cloud (https://rxresu.me)

When exporting to pdf the margins are now plain white instead of matching the background colour. 
This makes my resume look horrific. 

Kakuna

### Anything else?

Image

Also the random changes to the kakuna template were very annoying to fix.

martinfrancois:
## Summary
- set the HTML and body print background color to the resume background before PDF generation
- ensure Puppeteer margin areas inherit the same background color as the page design
- fix a bug where when content wrapped across pages it would not add the margin correctly at the bottom of the previous page

## Why
Issue #2741 reports PDF margins rendering as white for templates that apply margins via Puppeteer instead of CSS padding (for example Kakuna).

## Validation
- `pnpm typecheck`
- `pnpm exec biome check src/integrations/orpc/services/printer.ts`
- Created a lorem ipsum CV with dark background, for reference, here are the files created with the changes in this PR applied:
[lorem-ipsum_20260303_2305.json](https://github.com/user-attachments/files/25727836/lorem-ipsum_20260303_2305.json)
[lorem-ipsum_20260303_2305.pdf](https://github.com/user-attachments/files/25727837/lorem-ipsum_20260303_2305.pdf)

Fixes #2741

### DIFF
--- a/src/integrations/orpc/services/printer.ts
+++ b/src/integrations/orpc/services/printer.ts
@@ -180,7 +180,10 @@ export const printerService = {
 								pageContent.style.paddingLeft = `${pagePaddingX}pt`;
 								pageContent.style.paddingRight = `${pagePaddingX}pt`;
 							}
-							if (pagePaddingY > 0) pageContent.style.paddingTop = `${pagePaddingY}pt`;
+							if (pagePaddingY > 0) {
+								pageContent.style.paddingTop = `${pagePaddingY}pt`;
+								pageContent.style.paddingBottom = `${pagePaddingY}pt`;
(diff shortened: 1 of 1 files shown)


amruthpillai:
Good catch, will check this out locally and get this merged. Thank you so much for your help!

martinfrancois:
You're welcome @amruthpillai, thanks for the quick reaction! 😊

PR #2776, opened 2026-03-05
schema.ts update and migration file

Shepard246:


Shepard246:
I updated the schama.ts, created a migration file, and built the image, then de-reployed without changing the database to confirm proper migration. I only tested once with my setup, but this should at least help reduce your work some.

amruthpillai:
made a slightly similar change to this, but thank you for the PR!

PR #2761, opened 2026-03-01
feat(experience): add role progression to show career advancement within a company

TheDuke427:
**Problem**
The Experience section only supports a single position title and period per company entry. Users who were promoted or changed roles within the same company have no good way to represent this

**Solution**
This PR adds optional role progression to experience entries. When a user adds multiple roles to a single company entry, each role gets its own Position, Period, and Description. The company header stays unified with roles stacked beneath it.

Tested locally with the dev server. Created a test experience entry with 3 roles and verified role progression renders correctly on the resume preview.

- fixes #2721

PR #2777, opened 2026-03-05
fix(experience): remove bold from role titles and tighten role spacing

TheDuke427:
Following the merge of the role progression feature (#2761), two visual issues were present:

- Role position titles were rendering in bold (`<strong>` tag) making them visually heavier than intended
- Spacing between the company header and first role was too large due to the `mt-(--page-gap-y)` class

- Replaced `<strong>` with `<span>` for role position titles to match the weight of other metadata
- Changed role list top margin from `mt-(--page-gap-y)` to `mt-0` for tighter, more natural grouping under the company header

Tested locally on dev server

### DIFF
--- a/src/components/resume/shared/items/experience-item.tsx
+++ b/src/components/resume/shared/items/experience-item.tsx
@@ -45,11 +45,11 @@ export function ExperienceItem({ className, ...item }: ExperienceItemProps) {
 
 			{/* Role Progression */}
 			{hasRoles && (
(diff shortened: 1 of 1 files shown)


michlr:
but now all the roles in the one and only hospital I worked in, don't stand out anymore :-).
=> https://cv.michaelr.be/en
now I want the CSS section ba-ack....
('m using ° bullets - because the unicode ▪ symbols are not rendered anymore)

PR #2801, opened 2026-03-12
Make logged users able to view other public resumes

FrancocDev:
Hi. This PR is to fix the error that comes from see another public user resume when you are already logged. 
Thanks!


FrancocDev:
@amruthpillai Can you check this small change?

amruthpillai:
@FrancocDev looks good, nice catch on the query.

PR #2813, opened 2026-03-17
fix(onyx): restore profiles section to header right column

### LINKED ISSUE #2812: [Bug] Profiles used to render nicely in top right of header, but now they render below the header

### Product Variant

Cloud (https://rxresu.me)

Old nice rendering: 

Image

New broken rendering:

Image

Onyx

### Anything else?

_No response_

shaunporwal:
## Summary

Fixes #2812

In v5, the onyx header was refactored from a two-column grid to a single-row flexbox. This dropped the `profiles` section from the top-right of the header — profiles icons (LinkedIn, GitHub, etc.) either disappeared from the header entirely or rendered as a separate body section below it.

**Before (v3/v4 behavior, what users expect):**
Profiles icons appear in a right-aligned column in the header, alongside the name and contact info.

**After v5 refactor (broken):**
Profiles render as a body section below the header, or not at all.

## Fix

The `Header` component in `onyx.tsx` now reads `sections.profiles` from the resume store. When profiles is not hidden and has visible items, they are rendered in a right-aligned flex column on the same row as the name/contact info — restoring the original visual design.

- No changes to schema or other templates
- Fully backward compatible: if `profiles.hidden === true` or no visible items, the layout is unchanged (single column, same as current v5 behavior)
- Users who want profiles in the header should set `sections.profiles.hidden: false` and keep profiles out of their layout pages array to avoid duplication

## Changes

- `src/components/resume/templates/onyx.tsx` — `Header` component reads profiles from store and conditionally renders them right-aligned


amruthpillai:
I'll revisit this issue later, I'm hoping for a solution that could apply to all templates instead of making just the one template an edge case.

PR #2904, opened 2026-04-23
feat: allow hiding items in the left sidebar

Platinum1154:
### Summary
This PR adds the ability to hide items in the left sidebar, giving users more control over the layout and visibility of sidebar content.

### Changes
- add visibility toggle for sidebar items
- update UI to reflect hidden/shown state

### Motivation
In some cases, users may not need all sidebar items and would prefer a cleaner or more focused interface. This feature improves usability by allowing customization of visible content.

### Screenshots
image

image

### Testing
- verified that items can be hidden and shown correctly
- checked UI updates after toggling visibility
- ensured no regression in existing sidebar functionality


GoldenSection0618:
I tested this PR locally and found **two** behavior issues.

GoldenSection0618:
First, the existing resume section visibility toggle seems to be removed from the section dropdown. Previously, the Hide/Show action updated `draft.summary.hidden` or `draft.sections[type].hidden`, which controls whether a section appears in the resume itself, including preview/export behavior.

After this change, that entry point is removed and effectively replaced by a left sidebar visibility toggle. I think these are two different concepts and should remain separate:

- Resume section visibility: controls whether the section appears in the resume output.
- Sidebar item visibility: controls whether the section editor is shown in the left builder sidebar.

GoldenSection0618:
Second, the sidebar hidden state appears to be global. I notice when I hide a sidebar item in one resume, the same item is hidden when I open another resume. This seems to happen because the hidden state is persisted in `useSectionStore` without being keyed by `resumeId`.

If the intended behavior is per-resume sidebar customization, the hidden state should be scoped by `resumeId`.

amruthpillai:
I just don't think I see this as a direction the app should take. Collapsing a section is already possible, so hiding it seems like an unnecessary step. Thank you for your contribution and interest though.

PR #2906, opened 2026-04-23
feat: add OpenRouter as AI provider

iagodemacedo:
## Summary

- Adds [OpenRouter](https://openrouter.ai) as a new AI provider option, giving users access to hundreds of models (OpenAI, Anthropic, Google, Meta, Mistral, etc.) through a single API key.
- OpenRouter uses an OpenAI-compatible API, so this reuses the existing `@ai-sdk/openai` package with a custom base URL — **no new dependencies required**.

## Changes

- `src/integrations/ai/store.ts` — Added `"openrouter"` to the `AIProvider` type union.
- `src/integrations/orpc/services/ai.ts` — Added `"openrouter"` to the Zod provider schema and `getModel()` pattern match (uses `createOpenAI` with `https://openrouter.ai/api/v1` as base URL).
- `src/routes/dashboard/settings/ai.tsx` — Added OpenRouter to the provider combobox options in the settings UI.

## Test plan

- [ ] Select "OpenRouter" as the AI provider in Dashboard > Settings > AI
- [ ] Enter an OpenRouter API key and a model name (e.g., `anthropic/claude-sonnet-4`)
- [ ] Click "Test Connection" and verify it succeeds
- [ ] Enable AI features and test: chat, PDF/DOCX import, resume analysis, and resume tailoring
- [ ] Verify other providers still work as expected


GoldenSection0618:
I’m wondering whether `OpenRouter` is the right provider abstraction here.

The code is effectively using the OpenAI-compatible path:

```ts
createOpenAI({
  apiKey,
  baseURL: baseURL || "https://openrouter.ai/api/v1",
}).chat(model)
```
Since baseURL remains user-configurable, this is not strictly limited to OpenRouter. It could also work with other OpenAI-compatible endpoints.

Because of that, would it be clearer to expose this as an `OpenAI-compatible` provider instead, with OpenRouter used as the default/example Base URL?

GoldenSection0618:
A separate concern is model capability compatibility.

Reactive Resume’s AI features require different model capabilities, not just basic chat completion:

| Feature | Current implementation | Required model/API capability |
| --- | --- | --- |
| Test Connection | `generateText` with `Output.choice(...)` | Constrained output support |
| Resume Analysis | `generateText` with `Output.object(...)` | Reliable structured output / schema-constrained output |
| Tailor Resume | `generateText` with `Output.object(...)` | Reliable structured output / schema-constrained output |
| PDF/DOCX Import | Sends the uploaded document as a `file` part and expects valid resume JSON back | File/document input support + reliable JSON generation |
| AI Chat | `streamText` with the `patch_resume` tool | Streaming + tool calling support |

Because OpenRouter exposes many models with different capability levels, a model may pass the connection test but still fail on specific features, such as structured output or file-based resume import.

So the concern is not whether OpenRouter can work. It can. The concern is that users may assume any OpenRouter model works with all Reactive Resume AI features, while in practice different models support different capabilities. Without guidance, this could make the AI setup harder to use and harder to troubleshoot.

amruthpillai:
> A separate concern is model capability compatibility.
> 
> Reactive Resume’s AI features require different model capabilities, not just basic chat completion:
> 
> Feature	Current implementation	Required model/API capability
> Test Connection	`generateText` with `Output.choice(...)`	Constrained output support
> Resume Analysis	`generateText` with `Output.object(...)`	Reliable structured output / schema-constrained output
> Tailor Resume	`generateText` with `Output.object(...)`	Reliable structured output / schema-constrained output
> PDF/DOCX Import	Sends the uploaded document as a `file` part and expects valid resume JSON back	File/document input support + reliable JSON generation
> AI Chat	`streamText` with the `patch_resume` tool	Streaming + tool calling support
> Because OpenRouter exposes many models with different capability levels, a model may pass the connection test but still fail on specific features, such as structured output or file-based resume import.
> 
> So the concern is not whether OpenRouter can work. It can. The concern is that users may assume any OpenRouter model works with all Reactive Resume AI features, while in practice different models support different capabilities. Without guidance, this could make the AI setup harder to use and harder to troubleshoot.

This is true of most providers and models. It is, at the end of the day, up to the end user to decide on which model is the most optimal. Maybe through some research, we can provide suggested models for each of the ai providers and add it to this options schema in `src/routes/dashboard/settings/ai.tsx`.

PR #2966, opened 2026-05-06
feat: add AI observability via langchain and laminar

helder-mattos:
## Summary

Adds optional end-user configurable observability for AI workflows using Laminar and LangSmith.

## Why

AI-powered operations such as chat, resume analysis, resume tailoring, and document parsing can be difficult to debug without tracing. They can also be hard to monitor for latency, model behavior, and token usage.

This PR adds optional observability hooks so users can trace supported AI calls through Laminar or LangSmith when they choose to configure a provider.

## Changes

- Added AI observability configuration support.
- Added Laminar tracing integration for AI SDK calls.
- Added LangSmith tracing integration for AI SDK calls.
- Passed observability metadata through supported AI workflows.
- Added UI/settings support for configuring an observability provider.
- Kept observability disabled by default when no provider is configured.

## Notes

Observability is designed as an end-user setting, similar to AI provider configuration. It is not enabled globally for the whole platform through deployment-level environment variables.

When observability is not configured, existing AI workflows should continue to behave as before.


helder-mattos:
@amruthpillai Are you interested in this with the new base, or not at all? I really think that observability could be useful when using own API keys.

PR #2970, opened 2026-05-07
v5.1.0

amruthpillai:
**Self-hosters, please read before upgrading.**

  - **No more Browserless / Chromium dependency.** From v5.1.0 onwards, you do **not** need to run a `printer`, Browserless, or headless Chrome service alongside Reactive Resume. PDF generation is now performed entirely in your browser via `@react-pdf/renderer`. The `BROWSERLESS_TOKEN`, `PRINTER_ENDPOINT`, `PRINTER_APP_URL`, and `FLAG_DEBUG_PRINTER` environment variables are no longer read and can be removed from your `.env`.
  - **Database migrations.** This release ships several new schema migrations to bring the resume model up to date. They run automatically on container start, but you should **back up your PostgreSQL database before upgrading** in case rollback is needed.
  - **Custom CSS has been removed.** Because the resume PDF is no longer produced by a headless browser, raw CSS no longer applies to the exported document. A first-class template-customization story is on the roadmap for a future release.

## Highlights

- **PDF generation moved fully client-side.** `@react-pdf/renderer` now produces the exported PDF directly in your browser, and the in-builder live preview is rendered with `pdfjs`, so what you see in the preview is exactly what you'll download.
- **All resume templates re-ported.** Azurill, Bronzor, Chikorita, Ditgar, Ditto, Gengar, Glalie, Kakuna, Lapras, Leafish, Meowth, Onyx, Pikachu, and Rhyhorn have all been re-implemented on the new renderer with section-by-section parity.

## Resume Builder

- **Pixel-accurate live preview.** The builder preview is now rendered with `pdfjs` against the same document tree used for export, eliminating the long-standing "looks fine in the preview but breaks in the PDF" class of bugs.
- **TanStack Form everywhere.** All 35 form-bearing surfaces in the app — login, register, forgot-password, resume sections, settings, and dialogs — have been migrated to TanStack Form. `react-hook-form` and `@hookform/*` packages have been removed.
- **TanStack Query is the single source of truth for resume state.** The previous standalone Zustand resume store has been retired; optimistic updates, refetching, and error recovery now flow through one cache, behaving consistently across tabs and reloads.
- **Native page format and margins.** Page size and margin controls are now expressed directly to `@react-pdf/renderer`, replacing the Puppeteer-specific `printMarginTemplates` workaround.

## Self-Hosting

- **One-service deployment.** Self-hosted Compose files no longer need a `printer:` service block. The minimum stack is now just Reactive Resume + PostgreSQL (with optional S3-compatible storage and SMTP).
- **`/api/health` simplified.** The health endpoint now reports `database` and `storage` only, returning HTTP `503` if either dependency is unhealthy. There is no longer a `printer` field to monitor.
- **Updated documentation.** The Self-Hosting, Quickstart, Development, Architecture, and Legal pages have all been refreshed to describe the simplified single-service deployment.

## Authentication & AI

- **Better Auth retained end-to-end** with all nine plugins — passkeys, two-factor authentication, OAuth (Google / GitHub / LinkedIn / custom), API keys, dynamic OAuth client registration, dashboard, sentinel, and admin. Existing sessions and logins carry over after the upgrade.
- **AI features carry over** (Resume Analysis, AI Chat with JSON Patch tool calls, PDF / DOCX import) with the same provider matrix (OpenAI, Google Gemini, Anthropic, Ollama, OpenRouter). The AI router and service layer have been cleanly decoupled from the jobs / tailor flow for a more modular surface.

## Removed

- **Custom CSS option** in the resume builder. With rendering bound to `@react-pdf/renderer`, raw CSS no longer applies to the exported PDF. A structured customization API will return in a future release.
- **`PRINTER_*` and `BROWSERLESS_*` environment variables**, the `printer/{resumeId}` route, the `getByIdForPrinter` server function, and the printer-token signing code are all gone.

## Maintenance

- Removed dead printer-token utilities and the legacy public RPC surface for printer-only resume reads.
- Tightened the boundary between server-only and client-only code: server modules (database, storage, auth) are lazy-imported so they no longer leak into the client bundle.
- Reverse-proxy timeout guidance in the nginx example has been simplified now that no long-running PDF generation request flows through the proxy.

**Full Changelog**: [v5.0.20...v5.1.0](https://github.com/amruthpillai/reactive-resume/compare/v5.0.20...v5.1.0)


Dockerfile
```diff
@@ -46,20 +49,23 @@ LABEL org.opencontainers.image.url="https://rxresu.me"
...
 
-COPY --from=builder /app/.output ./.output
-COPY --from=builder /app/migrations ./migrations
-COPY --from=dependencies /tmp/prod/node_modules ./node_modules
+COPY --from=runtime-deps --chown=node:node /runtime-deps/node_modules ./node_modules
+COPY --from=builder --chown=node:node /app/apps/web/.output ./apps/web/.output
+COPY --from=pruner --chown=node:node /app/migrations ./migrations
 
-EXPOSE 3000/tcp
+WORKDIR /app/apps/web
```
stavros-k, line 63: This seem to cause issue with where local fs data are stored. 
https://github.com/amruthpillai/reactive-resume/blob/a43538c919f0b252b41334263b5b9681edfdef5b/packages/api/src/services/storage.ts#L117

cwd now resolves to /app/apps/web+/data instead of /app+/data
@amruthpillai

Issue #3006, opened 2026-05-08
[Bug] <The cloud resume displays Chinese characters with garbled characters.>


gitchenyiwu:
### Is there an existing issue for this?

- [x] Yes, I have searched the existing issues and none of them match my problem.

### Product Variant

Cloud (https://rxresu.me)

### Describe the bug you're experiencing

Last week, I saved a Chinese resume. But this morning, when I opened the resume to make changes, the Chinese characters became garbled. And when I opened a new document to input Chinese characters, there were also garbled characters or blank spaces. The information bar on the left can display Chinese, but the online resume is full of garbled characters.

<img width="1030" height="219" alt="Image" src="https://github.com/user-attachments/assets/e150273d-461f-4dee-9c8b-a0054262d141" />

### What template are you using?

None

### Anything else?

I tried different Chinese encoding methods, but none of them worked.

unipoco:
Same as #2986

amruthpillai:
This is fixed now, thanks to @JamesGoslings!

PR #3016, opened 2026-05-09
fix(pdf): register CJK fallback font so Chinese/Japanese/Korean text …

### LINKED ISSUE #2986: [Bug] Chinese characters garbled in v5.1.0

### Product Variant

Cloud (https://rxresu.me)

Version: v5.1.0 (Cloud)

Description:
Chinese characters render as garbled/missing glyphs in the resume preview and in exported PDF.

Steps to reproduce:
1. Open https://rxresu.me
2. Create or import a resume containing Chinese text
3. Preview the resume
4. Export the resume as PDF
5. Observe garbled text

Expected behavior:
Chinese text renders correctly in both preview and PDF export.

Actual behavior:
Chinese text becomes garbled/unreadable or missing.

Environment:
- OS: (please fill)
- Browser: (please fill)
- Font/Language settings: Chinese text present

Notes:
This may be related to font fallback or font embedding in v5.1.0.

None

### Anything else?

_No response_

JamesGoslings:
@amruthpillai 

## Summary

Fixes #2986. Restores correct rendering of Chinese / Japanese / Korean text in the PDF preview and export by registering a CJK-capable font as a glyph-level fallback alongside the user-selected typography.

| Before | After |
|---|---|
|image|image|

## Root cause

Since v5.1.0 the renderer was migrated from Puppeteer to `@react-pdf/renderer`. The new pipeline (`packages/pdf/src/hooks/use-register-fonts.ts`) only registers the user-selected typography family — Roboto, IBM Plex Serif, Open Sans, etc. — none of which contain CJK glyphs.

When the resume contains a CJK character:

1. The `<Text>` is styled with `fontFamily: 'Roboto'` (a single string).
2. `@react-pdf/renderer`'s textkit layer looks up each codepoint in Roboto.
3. Roboto has no glyph for the CJK codepoint → it falls back to `.notdef`.
4. The PDF is rendered with `.notdef` glyphs for every CJK character, which look like garbled boxes.

There was no CJK fallback registered, and `fontFamily` was always passed as a single string, so textkit's per-codepoint substitution path was never engaged.

## Fix

`@react-pdf/renderer`'s textkit layer **already supports per-codepoint font substitution** when a `<Text>` is styled with `fontFamily` as a **string array** — provided every family in the stack has actually been `Font.register()`-ed. This PR wires that up:

### 1. `packages/fonts` — pick the right CJK fallback

New helper `getPdfCjkFallbackFontFamily(family)`:

- Returns **`Noto Sans SC`** for sans-serif primary fonts, **`Noto Serif SC`** for serif primary fonts (matches the visual style).
- Returns **`null`** when no fallback is needed:
  - primary is a standard PDF font (Helvetica / Courier / Times) — Latin-only by design;
  - primary is already the resolved CJK fallback;
  - the resolved fallback isn't actually shipped as a webfont.
- Source Han Sans/Serif SC covers all CJK-Unified ideographs, so a single font handles **Simplified Chinese, Traditional Chinese, Japanese kanji, and Korean hanja** — i.e. all the locales reporting #2986 / #3006.

### 2. `packages/pdf/hooks/use-register-fonts` — register the fallback and widen the stack

After registering the primary body/heading fonts as before, additionally `Font.register(...)` the resolved CJK fallback. Only **regular weight** is registered for the fallback — substitution is per-codepoint, not per-weight, so a single face covers headings and body at any weight (one extra ~8 MB file, vs. registering all six weight/italic combinations).

The function's return type is widened to a new `PdfTypography` whose `body.fontFamily` and `heading.fontFamily` become a `[primary, cjkFallback]` two-element stack.

### 3. `packages/pdf/document` — propagate the widened type

A cast carries the widened runtime value back through `ResumeData` without changing the public `Typography` schema. All 15 templates already read `metadata.typography.body.fontFamily` directly, and `StyleSheet.fontFamily` accepts both `string` and `string[]`, so **no template edits are required**.

## Why this is safe for existing (Latin-only) users

- For standard PDF fonts (Helvetica/Courier/Times) and resumes whose primary already is a CJK font, `getPdfCjkFallbackFontFamily` returns `null`, so no extra `Font.register` call is made.
- When no fallback applies, `registerFonts` returns the **original typography reference unchanged** (zero allocation, identical to current behaviour).
- Even when the fallback is registered, textkit's `fontSubstitution` only consults it for codepoints the primary font cannot render. Latin glyphs continue to come from the primary font with identical metrics — a Roboto resume looks pixel-identical to before.

## Testing

- ✅ Reproduced the original bug on `main` with a Simplified Chinese resume (Azurill / Onyx / Chikorita templates) — characters rendered as garbled boxes.
- ✅ With this branch, the same resume renders correctly in both the live preview and the exported PDF.
- ✅ Existing Latin-only resumes render identically (verified visually against `main`).
- ✅ `pnpm lint` clean, `pnpm typecheck` clean for `packages/pdf` and `packages/fonts`.

## Known follow-up (out of scope for this PR)

While verifying the fix I noticed a separate, narrower issue: when a resume has typography **`lineHeight` set below ~1.35** _and_ contains CJK content, the descenders of CJK glyphs (e.g. 育, 实, 程) are visually clipped by the next line. Resumes at the default `lineHeight` of 1.5 are unaffected.

Root cause is independent of this PR: textkit computes the line box from the primary (Latin) font's ascent/descent, but Source Han Sans/Serif SC glyphs nearly fill the em-box, so a tightened line-height causes the glyph to extend past its line box. Browsers handle this transparently via mixed-script line-box rules, react-pdf does not.

This only manifests *because* CJK glyphs now render at all (i.e. it was masked by the original `.notdef` rendering). I will open a separate issue + PR for it once this one lands so the two changes can be reviewed independently.

## Notes

- The CJK font is downloaded lazily by `@react-pdf/renderer` on first render that needs it, so users who never have CJK content pay no bandwidth or memory cost.

### DIFF
--- a/packages/fonts/src/index.ts
+++ b/packages/fonts/src/index.ts
@@ -184,6 +184,25 @@ export function getFallbackWebFontFamilies(family: string) {
 	return fallback === family ? [] : [fallback];
 }
 
+/**
+ * Returns a CJK web font (Noto Sans/Serif SC) to register as a glyph-level
+ * fallback for PDF rendering, or `null` when no fallback is needed
+ * (standard PDF font, or primary already is the fallback).
+ *
+ * Source Han Sans/Serif SC covers all CJK-Unified ideographs, so a single
+ * font handles Simplified/Traditional Chinese, Japanese kanji and Korean
+ * hanja — the locales reporting #2986 / #3006.
+ */
+export function getPdfCjkFallbackFontFamily(family: string): string | null {
+	if (isStandardPdfFontFamily(family)) return null;
+
+	const fallback = getPrimaryCjkWebFont(family);
+	if (fallback === family) return null;
+	if (!getWebFont(fallback)) return null;
+
+	return fallback;
+}
+
 export function getLoadableWebFontWeights(family: string, preferredWeights: string[]) {
 	const font = webFontMap.get(family);
 	if (!font) return [];
--- a/packages/pdf/src/document.tsx
+++ b/packages/pdf/src/document.tsx
@@ -1,4 +1,4 @@
-import type { LayoutPage, ResumeData } from "@reactive-resume/schema/resume/data";
+import type { LayoutPage, ResumeData, Typography } from "@reactive-resume/schema/resume/data";
 import type { Template } from "@reactive-resume/schema/templates";
 import type { ComponentType } from "react";
 import type { SectionTitleResolver } from "./section-title";
@@ -23,8 +23,13 @@ export type ResumeDocumentProps = {
 export const ResumeDocument = ({ data, template, resolveSectionTitle }: ResumeDocumentProps) => {
 	const TemplatePageComponent = getTemplatePage(template);
 	const typography = registerFonts(data.metadata.typography);
+	// `registerFonts` widens `fontFamily` to `string | string[]` for CJK
+	// fallback (#2986); the cast carries that wider runtime value through
+	// `ResumeData` without changing the public schema.
 	const resumeData =
-		typography === data.metadata.typography ? data : { ...data, metadata: { ...data.metadata, typography } };
+		typography === data.metadata.typography
+			? data
+			: { ...data, metadata: { ...data.metadata, typography: typography as unknown as Typography } };
 
 	return (
 		<RenderProvider data={resumeData} resolveSectionTitle={resolveSectionTitle}>
--- a/packages/pdf/src/hooks/use-register-fonts.ts
+++ b/packages/pdf/src/hooks/use-register-fonts.ts
@@ -1,7 +1,12 @@
 import type { FontWeight } from "@reactive-resume/fonts";
 import type { Typography } from "@reactive-resume/schema/resume/data";
 import { Font } from "@react-pdf/renderer";
-import { getFont, getWebFontSource, isStandardPdfFontFamily } from "@reactive-resume/fonts";
+import {
+	getFont,
+	getPdfCjkFallbackFontFamily,
+	getWebFontSource,
+	isStandardPdfFontFamily,
+} from "@reactive-resume/fonts";
 
 type FontWeightRange = {
 	lowest: number;
@@ -10,6 +15,13 @@ type FontWeightRange = {
 
 const registeredFontVariants = new Set<string>();
 const fallbackFontFamily = "IBM Plex Serif";
+
+// `fontFamily` is widened to `string | string[]` so react-pdf can do
+// glyph-level font fallback for CJK characters (#2986).
+export type PdfTypography = Omit<Typography, "body" | "heading"> & {
+	body: Omit<Typography["body"], "fontFamily"> & { fontFamily: string | string[] };
+	heading: Omit<Typography["heading"], "fontFamily"> & { fontFamily: string | string[] };
+};
 
 const getFontWeightRange = (fontWeights: string[]): FontWeightRange => {
 	const numericWeights = fontWeights.map(Number).filter((weight) => Number.isFinite(weight));
@@ -53,7 +65,7 @@ const resolvePdfTypography = (typography: Typography): Typography => {
 	};
 };
 
-export const registerFonts = (typography: Typography): Typography => {
+export const registerFonts = (typography: Typography): PdfTypography => {
 	Font.registerHyphenationCallback((word) => [word]);
 
 	const pdfTypography = resolvePdfTypography(typography);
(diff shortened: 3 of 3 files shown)


amruthpillai (commented):
# Code Review

## Overview

Fixes #2986 (CJK characters rendering as `.notdef` boxes in PDF preview/export). The PR registers a CJK-capable webfont (Noto Sans SC / Noto Serif SC) as a glyph-level fallback in `@react-pdf/renderer`'s textkit layer, and widens the body/heading `fontFamily` to a `[primary, cjkFallback]` array so per-codepoint substitution kicks in. Three files, +68/-5.

## Code quality

The implementation is small, surgical, and well-justified. A few things I particularly like:

- **Reference-stability preserved on the Latin-only path** (`use-register-fonts.ts:111-113`). The `if (!bodyCjkFallback && !headingCjkFallback) return pdfTypography as PdfTypography;` short-circuit keeps `typography === data.metadata.typography` identity intact in `document.tsx:26`, so existing Latin resumes get the exact same allocation behavior as before.
- **Schema-preserving type widening.** `PdfTypography` is local to the PDF package and the `as unknown as Typography` cast is documented with a comment pointing back at #2986. Avoiding a public schema change is the right call — `StyleSheet.fontFamily` already accepts `string | string[]` so templates need no edits.
- **Single-weight CJK registration is the correct trade-off.** Source Han Sans/Serif SC is ~8 MB per face; loading 6 weights would balloon the bundle for what textkit treats as per-codepoint substitution anyway.
- **Lazy fetch claim holds.** `@react-pdf/renderer`'s font loader is glyph-driven, so registering the fallback eagerly does not download bytes for Latin-only resumes — confirms the "zero bandwidth cost" claim in the PR description.

## Issues / suggestions

**1. Standard PDF font users with CJK content still get broken output (medium).**
`getPdfCjkFallbackFontFamily` returns `null` when the primary is Helvetica/Courier/Times-Roman (`packages/fonts/src/index.ts:198`). The PR rationalizes this as "Latin-only by design," but functionally a user who picks Helvetica + writes Chinese will still see `.notdef` boxes — the same bug #2986 is fixing, just for a smaller cohort. `StyleSheet.fontFamily` accepts a `[Helvetica, "Noto Sans SC"]` stack and textkit handles it, so the early-return on `isStandardPdfFontFamily` could be lifted. If you want to keep this out of scope, please call it out explicitly in the PR description as a known limitation rather than "by design."

**2. CJK glyphs always render at weight 400, even in headings (low — worth documenting).**
Because only `(fallback, 400, regular)` is registered, a heading with weight 700 will render its CJK characters at weight 400. The PR description correctly notes "substitution is per-codepoint, not per-weight," but the user-visible consequence is that a bolded heading mixing Latin and CJK will render with bold Latin and regular-weight CJK side by side. This is the right cost/benefit choice (vs. a 6× bundle increase), but worth a one-line note in the "known follow-up" section so reviewers/users aren't surprised.

**3. `getPrimaryCjkWebFont(family) === family` is a weak guard (low).**
`getPrimaryCjkWebFont` only ever returns `"Noto Sans SC"` or `"Noto Serif SC"`, so the `fallback === family` check only short-circuits for those two exact strings. Other Chinese-capable webfonts shipped in `webFontList` (e.g., `ZCOOL QingKe HuangYou`, listed in `preferredChineseFontFamilies`) will get Noto Sans SC stacked on top. That's harmless (textkit only consults the fallback for codepoints the primary lacks), but it does cause an unnecessary `Font.register` call and ~8 MB lazy fetch trigger if the user types a glyph the display font is missing. Consider checking against a small `Set` of "already covers CJK Unified ideographs" families instead.

**4. The `headingCjkFallback !== bodyCjkFallback` guard is redundant (nit).**
At `use-register-fonts.ts:107`, the `!== bodyCjkFallback` check duplicates the dedup that `registeredFontVariants` already does inside `registerFont`. Not wrong, just dead weight — drop the second clause for clarity, or keep it if you prefer the explicit shape.

**5. Cast in `document.tsx` could be tightened (nit).**
The `typography as unknown as Typography` double-cast at `document.tsx:31` works but is a code smell. A cleaner option is to make `ResumeData["metadata"]["typography"]` accept the wider type via a local interface where it's consumed, but that's a larger refactor. Acceptable as-is given the comment.

## Test coverage

No automated tests added. The package has no pre-existing tests for `use-register-fonts` so this isn't out of step with conventions, but the typography object's reference-equality contract is now load-bearing (downstream identity check in `document.tsx`). A small unit test asserting:

- Latin-only input → returned object is `===` identical to the resolved typography
- CJK-eligible input → `body.fontFamily` is `[primary, "Noto Sans SC"]`
- Helvetica → no fallback registered (currently — see issue 1)

would lock the contract in for future refactors. Not a blocker.

## Security / performance

- **No security concerns.** All font sources come from the existing `webfontlist.json`; no new network surface.
- **Performance:** the only new work on the Latin path is two `getPdfCjkFallbackFontFamily` calls (each O(1) map lookups). Negligible.
- **Bundle size:** unchanged for Latin-only users; ~8 MB additional lazy fetch on first CJK render, as documented.

## Verdict

Nice fix — correct, scoped, and the architectural choice (widen at the PDF-package boundary, leave the schema alone) is good. The most actionable feedback is item #1 — either extend the fallback to standard PDF fonts or document that limitation explicitly. Items #2–5 are polish.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #3044, opened 2026-05-12
fix(pdf): prevent empty Profiles background bleed in Ditgar and Gengar templates

### LINKED ISSUE #3043: [Bug] Ditgar/Gengar with no profiles have background bleed into Expierences

### Product Variant

Cloud (https://rxresu.me)

Using the Ditgar or Gengar layouts, without a Profile defined, has the top background bleed into the Experiences section.

Image

https://rxresu.me/swbusche/test-for-bug

Ditgar

### Anything else?

The sample image, https://rxresu.me/templates/jpg/gengar.jpg, looks like the Summary would be displayed there, but only by adding a Profile, could I get the background to render correctly again.

its-rajesh-smp:
Closes #3043

## What was broken

When using the Ditgar or Gengar templates without any visible Profiles items, the first visible main section was promoted into the special top section.

For resumes where `profiles` was configured first but had no visible items, `experience` could move into the tinted header/profile area, causing the profile background color to bleed into the Experience section.

## Root cause

Both templates filtered the main section list first and then used the first remaining visible section as the special top section.

That meant an empty configured first section, such as `profiles`, was removed before the template decided which section should receive the special top treatment. The next visible section, such as `experience`, was then incorrectly rendered in that tinted container.

## Changes

`packages/pdf/src/templates/shared/filtering.ts`

- Added `splitFirstVisibleSection`, which keeps the configured first section separate only when that original first section is visible.
- Remaining sections are still filtered normally.

`packages/pdf/src/templates/ditgar/DitgarPage.tsx`

- Uses `splitFirstVisibleSection` for first-page main sections.
- Prevents `experience` or other later sections from being promoted when `profiles` is empty.

`packages/pdf/src/templates/gengar/GengarPage.tsx`

- Applies the same first-section behavior as Ditgar.

`packages/pdf/src/templates/shared/filtering.test.ts`

- Added regression coverage for the empty `profiles` case.
- Added coverage for the visible first-section case.

## Testing

- [ ] Ditgar template with no Profiles — Experience section background should NOT bleed into profile area
- [ ] Gengar template with no Profiles — same background bleed check
- [ ] Ditgar template WITH Profiles defined — layout renders correctly as before
- [ ] Gengar template WITH Profiles defined — layout renders correctly as before
- [ ] First visible section is promoted correctly when Profiles is empty
- [ ] PDF output matches expected visual for both templates

### DIFF
--- a/packages/pdf/src/templates/ditgar/DitgarPage.tsx
+++ b/packages/pdf/src/templates/ditgar/DitgarPage.tsx
@@ -7,7 +7,7 @@ import { parseColorString, rgbaStringToHex } from "@reactive-resume/utils/color"
 import { useRender } from "../../context";
 import { CustomFieldContactItem, WebsiteContactItem } from "../shared/contact-item";
 import { TemplateProvider } from "../shared/context";
-import { filterSections } from "../shared/filtering";
+import { filterSections, splitFirstVisibleSection } from "../shared/filtering";
 import { getTemplateMetrics } from "../shared/metrics";
 import { getTemplatePageMinHeightStyle, getTemplatePageSize } from "../shared/page-size";
 import { hasTemplatePicture } from "../shared/picture";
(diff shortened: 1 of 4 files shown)


amruthpillai:
@its-rajesh-smp This issue should be fixed as of the latest update, thank you for your contribution and time spent on the project. It is very much appreciated!

PR #3070, opened 2026-05-15
fix(pdf): align textkit line-box and font metrics to browser behaviour

### LINKED ISSUE #3069: [Bug] CJK characters are clipped at the bottom when typography line-height is below ~1.4

### Product Variant

Cloud (https://rxresu.me)

@amruthpillai 
## Summary

When a resume contains CJK (Chinese / Japanese / Korean) text **and** the typography `lineHeight` is set below ~1.4, the bottom of CJK glyphs (descenders, and the lower portion of common ideographs like 育, 实, 程) is clipped by the following line. Latin characters in the same resume render correctly. Resumes left at the default `lineHeight` of 1.5 are unaffected.

This was masked by #2986 (CJK characters rendered as `.notdef` boxes); once #3013 restored real CJK rendering, the underlying line-box bug surfaced.

## Reproduction

1. Create or open a resume with Simplified Chinese content.
2. Settings → Typography → set Body Line Height to **1.15** or **1.2** (any value below ~1.35 reproduces).
3. Inspect the live PDF preview / exported PDF.

**Expected**: CJK glyphs render fully — same density as the v5.0.x Puppeteer renderer produced.
**Actual**: The bottom ~10–15 % of every line of CJK text is cut off by the next line.

### Screenshots

**Current behaviour on `main` (v5.1.4)** — descenders clipped:

Image

**Expected (v5.0.x Puppeteer renderer, same resume JSON)** — full glyphs:

Image

**Closeup of a single line** — easier to see the clipping:

> _<paste a tightly cropped screenshot showing the descender of e.g. 育 / 实 / 程 being cut by the next line>_

Tested on `main` (`v5.1.4`), Azurill / Onyx / Meowth templates. Same behaviour expected for Traditional Chinese, Japanese kanji and Korean hanja since they share the Source Han glyph metrics.

## Root cause

The bug is two independent gaps in `@react-pdf/textkit` relative to standard CSS line-box rules. Both must be fixed together; fixing only one either leaves the clipping or makes layout balloon.

### 1. textkit's `height(run)` short-circuits user lineHeight

`@react-pdf/textkit/lib/textkit.js`:

```js
const height = (run) => {
  const lineHeight = run.attributes?.lineHeight;
  return lineHeight || lineGap(run) + ascent(run) - descent(run);
};
```

CSS specifies a line-box height of `max(line-height, content-area)`. textkit's `||` discards the content-area whenever a `lineHeight` is set. The line baseline is still computed from the run's real ascent (which **does** account for the CJK font), so the descender is positioned outside the (too-short) box and ends up overlapping the next line.

### 2. textkit reads hhea metrics, not OS/2 typo metrics

`ascent`, `descent` and `lineGap` come straight from fontkit's hhea defaults. For Source Han Sans / Serif (used as the CJK fallback registered in #3013) the hhea ascender is intentionally inflated for legacy Windows GDI compatibility:

```
Noto Sans SC (1000 unitsPerEm):
  hhea ascent / descent       =  1160 / -288  →  intrinsic 1.448 em   ← textkit
  OS/2 typoAscender/Descender =   880 / -120  →  intrinsic 1.000 em   ← Chromium / v5.0.x
                                                  ~31 % difference
```

Browsers (including Chromium, which v5.0.x's Puppeteer renderer used) prefer the OS/2 typo triple. textkit always picks the larger hhea values, so even a properly enforced `Math.max` would over-inflate the line-box for CJK content.

For comparison:

| Font | hhea | OS/2 typo | Affected by switching? |
|---|---|---|---|
| Helvetica / Courier / Times-Roman (no OS/2 table) | hhea | n/a | No (fallback path) |
| IBM Plex Serif | 1.300 em | 1.300 em | No |
| Roboto | 1.172 em | 1.000 em | Only if `lineHeight < 1.17` |
| Noto Sans SC | **1.448 em** | **1.000 em** | Yes, this is the fix |

## Why v5.0.x didn't have this

Puppeteer = Chromium = uses OS/2 typo metrics + applies CSS `max()` line-box rules. Both gaps disappear; CJK content lays out exactly as the type designers intended.

## Proposed fix

Two minimal changes in `@react-pdf/textkit`:

```js
// resolveTypoMetrics(font): prefer OS/2 sTypoAscender/Descender/LineGap
//   when present, fall back to fontkit's hhea defaults.
// ascent/descent/lineGap helpers go through resolveTypoMetrics.

const height = (run) => {
  const lineHeight = run.attributes?.lineHeight;
  const intrinsic = lineGap(run) + ascent(run) - descent(run);
  return Math.max(lineHeight || 0, intrinsic); // CSS line-box rule
};
```

Total ~30 LOC. I have a working version locally.

## I'm working on this

I'll open a PR shortly that ships this as a `pnpm patch` against `@react-pdf/textkit@6.3.0` (so existing users are unblocked immediately), and a separate PR upstream at `diegomura/react-pdf` so the patch can eventually be removed.

---

Refs: #2986, #3013 (CJK fallback registration), #2989 (legacy local font aliasing)

None

### Anything else?

_No response_

JamesGoslings:
Closes #3069.

## TL;DR

CJK characters in resumes with line-height set below ~1.4 had their descenders clipped by the next line. Two independent gaps in `@react-pdf/textkit` vs CSS line-box rules cause this. Ships a `pnpm patch` (~30 LOC) that aligns textkit with browser behaviour.

| | Before (`main` v5.1.4) | After (this PR) |
|---|---|---|
| Preview / PDF | image | image |

## What's wrong

See #3069 for the full explanation. The two bugs are:

1. **`height(run)` short-circuits `lineHeight`** — CSS specifies line-box height = `max(line-height, content-area)`, but textkit treats user `lineHeight` as authoritative. Combined with the real (CJK-aware) baseline calculation, descenders fall outside the box.

2. **textkit reads hhea metrics, not OS/2 typo metrics** — for Source Han Sans / Serif (the CJK fallback registered in #3013) hhea is inflated to 1.45 em vs typo's 1.0 em, for legacy Windows GDI compatibility. Chromium uses typo metrics; textkit doesn't.

## Fix

Patches `@react-pdf/textkit@6.3.0` via pnpm:

```js
// New helper, used by ascent / descent / lineGap.
const resolveTypoMetrics = (font) => {
  const os2 = font?.['OS/2'];
  if (!os2 || typeof os2.typoAscender !== 'number' /* … */)
    return { ascent: font.ascent, descent: font.descent, lineGap: font.lineGap };
  return {
    ascent: os2.typoAscender,
    descent: os2.typoDescender,
    lineGap: os2.typoLineGap ?? font.lineGap ?? 0,
  };
};

// CSS line-box rule.
const height = (run) => {
  const lineHeight = run.attributes?.lineHeight;
  const intrinsic = lineGap(run) + ascent(run) - descent(run);
  return Math.max(lineHeight || 0, intrinsic);
};
```

The OS/2 fallback to hhea preserves behaviour for the StandardFont stand-ins (Helvetica / Courier / Times-Roman) that don't carry an OS/2 table, and for any web font where the two tables agree.

## Compatibility

| Scenario | Effect |
|---|---|
| Latin-only resume on Helvetica / IBM Plex Serif | None |
| Latin-only resume on Roboto, lineHeight ≥ 1.17 | None |
| Latin-only resume on Roboto, lineHeight < 1.17 | Slightly tighter (now matches Chromium) |
| CJK resume, lineHeight ≥ 1.5 | None |
| CJK resume, lineHeight < ~1.4 | **Fixed**: lays out like the v5.0.x renderer did |

## Tooling

- **`Dockerfile.dev`**: added `COPY patches ./patches` before `pnpm install` so the dev image build doesn't fail on `--frozen-lockfile`. The reference in `package.json`'s `pnpm.patchedDependencies` requires the patch file at install time.
- **`Dockerfile`** (production): unchanged — `turbo prune --docker` already includes `patches/` in its output (verified locally; the patch reference makes it part of the pruned slice).
- **`autofix.yml`** and other CI workflows: unchanged. They run `pnpm install` from the repo root where `patches/` lives.

## Upstream

Shipped as a patch because users see the bug today; an equivalent PR will be opened on `diegomura/react-pdf` so the patch can be removed once a fixed `@react-pdf/textkit` is published. The patch entry can then be dropped from `pnpm.patchedDependencies` and the file deleted in a one-line follow-up.

## Testing

- ✅ Reproduced on `main` with a Simplified Chinese resume across Azurill / Onyx / Meowth templates.
- ✅ With this branch: clipping resolved; line density visually matches v5.0.x for the same `lineHeight=1.15` resume.
- ✅ Latin-only resumes on Roboto / IBM Plex Serif / Helvetica unchanged at typical line-heights.
- ✅ `pnpm install --frozen-lockfile` succeeds in a fresh checkout.
- ✅ `turbo prune web --docker` includes `patches/` automatically — production Dockerfile build path verified locally.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved PDF text layout by refining vertical font metrics and line-height calculations for more accurate rendering.

* **Chores**
  * Applied a targeted package patch and updated workspace/build config to include it.
  * Bumped several dependency and dev-dependency versions across the project for tooling and runtime updates.

[Review Change Stack](https://app.coderabbit.ai/change-stack/amruthpillai/reactive-resume/pull/3070?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


amruthpillai:
@JamesGoslings This is genius, also, I've never used git patches like this before, quite interesting. Thank you so much for the PR, will test it out locally and merge soon.

Issue #3085, opened 2026-05-19
Login Issue


abhimanyusingh18101998-del:
When i am getting my ID from google the error is shown and does not login and i have saved many resumes in my login

shreyasD01:
Yes I am not able to login into my old account, plus not able to sign up using fresh google account as well.

bioinformatist:
Do you login with GitHub acc? If so, you can try as: https://github.com/amruthpillai/reactive-resume/pull/3095#issuecomment-4570293183

amruthpillai:
> *This was generated by AI during triage.*

Thanks for reporting this. This looks related to the current cloud sign-in incident tracked in #3118, but this thread still needs a bit more detail to tell whether it is the same outage or a separate migrated-account issue.

Could you confirm:
- whether the affected account signs in with Google, GitHub, or email/password
- the exact error or redirect you see now, for example `signup_disabled` or `account_not_linked`
- whether this is still reproducible today on https://rxresu.me

If this is a GitHub-linked migrated account, the temporary workaround described here may help in the meantime: https://github.com/amruthpillai/reactive-resume/pull/3095#issuecomment-4570293183

AnnemijnWitkam:
I'm also having issues with logging in with Google. I keep getting `account_not_linked`. I have tried both logging in and signing up, either way I get the same `account_not_linked` error in my url.

tusharmotwani1718:
@AnnemijnWitkam If you are getting account_not_linked error probably the reason is you have created your account with email and password and now trying to log in with google or some other oauth provider.

Issue #3078, opened 2026-05-17
[Bug] <title>login issue


Avinash61160:
### Is there an existing issue for this?

- [x] Yes, I have searched the existing issues and none of them match my problem.

### Product Variant

Cloud (https://rxresu.me)

### Describe the bug you're experiencing

tried logging in but even after login it lands on the same page again for login

### What template are you using?

None

amruthpillai:
Thanks for the report. Before we can continue triage, please update this issue with the following information:

Provide the authentication method, current version and browser, exact redirect or error, and whether the account previously worked.

Once you add it, comment here and a maintainer can review the issue again.

Issue #3046, opened 2026-05-12
[Bug] <title>unable to connect using google or GitHub


sambalawbe:
### Is there an existing issue for this?

- [x] Yes, I have searched the existing issues and none of them match my problem.

### Product Variant

Cloud (https://rxresu.me)

### Describe the bug you're experiencing

I'm not able to sign in using Google or GitHub

### What template are you using?

None

zeyad-alasaad:
+1

amruthpillai:
Thanks for the report. Before we can continue triage, please update this issue with the following information:

Provide the affected sign-in method, current version and browser, exact error, approximate timestamp, and whether this affects a new or existing account.

Once you add it, comment here and a maintainer can review the issue again.

PR #3095, opened 2026-05-21
fix(auth): reconcile migrated social login accounts

bioinformatist:
## TL;DR

This PR fixes a migrated social-login reconciliation path without relaxing sign-up or account-linking policy.

Recent reports (#3046, #3078 and #3085) mention Cloud social login failures, and in my situation, my existing GitHub-associated account could still be redirected to `https://rxresu.me/?error=signup_disabled`. This can happen when Better Auth cannot reconcile the incoming OAuth profile with an existing migrated account, then treats the OAuth callback as a new signup while signups are disabled.

## Root Cause

The current social providers still intentionally disable implicit signup:

https://github.com/amruthpillai/reactive-resume/blob/39e88dd365e441eea7fac11169db66a632a5dd0a/packages/auth/src/config.ts#L340-L363

That policy is correct and this PR does not change it.

The problem is in the migrated-account lookup data. During migration, non-credential accounts were stored with
`account.accountId = originalUser.id`, not the upstream provider's account id:

https://github.com/amruthpillai/reactive-resume/blob/35e2daa8079bd8ff1818a7bcd14e3a6734191190/scripts/migration/user.ts#L359-L368

So for a migrated GitHub user, Better Auth may receive GitHub's current numeric profile `id`, but the stored `github`
account row can still have the old Reactive Resume user id as `account.accountId`.

This was previously handled in #2620 by looking up the legacy GitHub account by `username` / `displayUsername` and
returning the stored `accountId` as the mapped OAuth profile `id`:

https://github.com/amruthpillai/reactive-resume/blob/18e8aadf18bb5eaa126592350458b1f4fb81f7b0/src/integrations/auth/config.ts#L164-L194

Later, #2874 added useful case-insensitive email reconciliation for OAuth users:

https://github.com/amruthpillai/reactive-resume/blob/61b33249413ab97801a31eb97a2626d85c225c5b/src/integrations/auth/config.ts#L171-L185

However, that shared mapper no longer preserved the GitHub-specific legacy `accountId` reconciliation from #2620.

## What Changed

This PR extracts the OAuth profile mapping logic into `packages/auth/src/oauth-profile.ts` so it can be tested directly.

For all social providers, the mapper keeps the case-insensitive email lookup behavior and normalizes the matched stored
email before handing it back to Better Auth. This prevents a matched mixed-case database email from being passed back into
a later lower-cased lookup.

For GitHub only, the mapper restores legacy migrated-account reconciliation, but with a narrower condition than the old
#2620 logic:

- `account.providerId` must be `github`
- GitHub `profile.login` must match the existing user's `username` or `displayUsername`
- the OAuth email must match the existing user's email after normalization

If those conditions are met, the mapper returns the stored `account.accountId` as `id`, so Better Auth can find the
existing linked account instead of treating the callback as a new signup.

The email requirement is intentional. GitHub usernames can be changed and old usernames can become available again, so
this PR avoids a username-only account rescue path.

This PR does not change `disableSignUp`, `disableImplicitSignUp`, or `account.accountLinking`.

## Tests

I added focused unit tests for the mapper behavior:

- reuses a migrated GitHub account when `login` and normalized email match
- does not reuse a migrated GitHub account when only `login` matches
- falls back to normalized email matching when no legacy GitHub account matches
- normalizes matched existing-user emails before returning them to Better Auth
- avoids unnecessary DB updates when email is already normalized
- preserves new social-user username allocation behavior
- rejects provider profiles without email

Validation was run on NixOS through the repository `direnv` dev shell:

```bash
direnv exec . pnpm --filter @reactive-resume/auth test
direnv exec . pnpm --filter @reactive-resume/auth typecheck
direnv exec . pnpm exec biome check packages/auth/src/config.ts packages/auth/src/oauth-profile.ts packages/auth/src/oauth-profile.test.ts
git diff --check
```

Result: auth tests passed (`3` files, `12` tests), typecheck passed, Biome passed, and `git diff --check` passed.

## Delivery Testing Note

I did not do a production-like delivery test because the failing path depends on production OAuth credentials and migrated
account rows whose stored `account.accountId` differs from the provider's current OAuth id. A fresh local Docker deployment
would mostly test new accounts and fresh OAuth links, not the historical migration state that triggers this bug.

The unit tests cover the exact mapper contract that feeds Better Auth's OAuth account lookup, and the PR keeps the existing signup/linking policy unchanged.

---

Thank you for maintaining Reactive Resume. It has helped many people create better resumes and reach interview and job
opportunities, including me :heart:

## Release Notes

* **New Features**
  * Improved OAuth and GitHub social login with better handling of account information, legacy account migration, and automatic username generation for new social users.

* **Tests**
  * Added comprehensive test suite validating OAuth profile mapping, legacy account migration, email normalization, and error handling for invalid provider data.

[Review Change Stack](https://app.coderabbit.ai/change-stack/amruthpillai/reactive-resume/pull/3095?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/auth/src/config.ts
+++ b/packages/auth/src/config.ts
@@ -1,6 +1,4 @@
 import type { GenericOAuthConfig } from "better-auth/plugins";
-import type { SQL } from "drizzle-orm";
-import type { AnyPgColumn } from "drizzle-orm/pg-core";
 import type { JWTPayload } from "jose";
 import { apiKey } from "@better-auth/api-key";
 import { drizzleAdapter } from "@better-auth/drizzle-adapter";
@@ -8,14 +6,13 @@ import { dash } from "@better-auth/infra";
 import { oauthProvider } from "@better-auth/oauth-provider";
 import { passkey } from "@better-auth/passkey";
 import { compare, hash } from "bcrypt";
-import { APIError, BetterAuthError, betterAuth } from "better-auth";
+import { APIError, betterAuth } from "better-auth";
 import { createAuthMiddleware } from "better-auth/api";
 import { verifyAccessToken } from "better-auth/oauth2";
 import { admin, jwt } from "better-auth/plugins";
 import { genericOAuth } from "better-auth/plugins/generic-oauth";
 import { twoFactor } from "better-auth/plugins/two-factor";
 import { username } from "better-auth/plugins/username";
-import { eq, or, sql } from "drizzle-orm";
 import { createElement } from "react";
 import { db } from "@reactive-resume/db/client";
 import * as schema from "@reactive-resume/db/schema";
@@ -25,6 +22,7 @@ import { env } from "@reactive-resume/env/server";
 import { rateLimitConfig, TRUSTED_IP_HEADERS } from "@reactive-resume/utils/rate-limit";
 import { generateId, toUsername } from "@reactive-resume/utils/string";
 import { isAllowedOAuthRedirectUri } from "@reactive-resume/utils/url-security.node";
+import { createGithubProfileMapper, createProfileMapper } from "./oauth-profile";
 import { getTrustedOrigins } from "./trusted-origins";
 
 const authBaseUrl = env.APP_URL;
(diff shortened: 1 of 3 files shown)


AdamWHY2K:
I see this has been released in v5.1.6 and the cloud version is running the latest version, but I'm still experiencing this issue.

Finslove:
Same issue here. I've also tried clearing browser cache/cookies and testing in different browsers/incognito mode with no success. My account was registered with Google login before the migration. Since the fix was deployed 3 days ago (v5.1.6 on May 25), the server should be updated by now, but I'm still getting the account_not_linked error. It seems like some migrated accounts may need manual reconciliation. Could the team look into accounts that are still affected after the v5.1.6 deployment?

bioinformatist:
@AdamWHY2K Thanks for reporting this.

I ran into the same problem after `v5.1.6` was deployed. In my case, the temporary workaround was to set my GitHub
**public email** back to the email address that my Reactive Resume account originally used, then try GitHub login again.

One subtle detail: changing the GitHub **primary email** alone may not be enough. The GitHub OAuth profile can expose the
account's public profile email first, so if the public email differs from the email stored in Reactive Resume, the
migrated-account reconciliation in this PR may still not match.

So for GitHub-linked migrated accounts, a possible temporary workaround is:

1. Go to GitHub profile settings.
2. Set the **public email**(displayed in your GitHub **Profile** page) back to the email **previously used** by the Reactive Resume account.
3. Try signing in to `rxresu.me` with GitHub again.
4. After getting back into the account, avoid changing it until the account has been reconciled server-side.

This worked for me, but it is only a workaround. And not sure for Google-linked accounts like @Finslove mentioned.

@amruthpillai would you be open to a one-time/manual reconciliation step for affected migrated accounts? My current
understanding is that this PR safely fixes the cases where the incoming OAuth profile still matches the migrated
`user.email` and, for GitHub, `username` / `displayUsername`. But accounts whose OAuth email has drifted since migration
may still need an admin-side repair, for example updating/linking the stored `account.accountId` after confirming the user
identity from production data.

AdamWHY2K:
Yep, changing my public email back fixed for me. Thank you <3

ASHMITA-DE:
I am still unable to login with my credentials.

bioinformatist:
Yep, the workaround only works with GitHub login accounts. Sorry to hear that.

ASHMITA-DE:
Yeah but what about users who had logged in via email? @amruthpillai any resolution for that?

fumbles:
I actually was able to get around this. I reset my password successfully. Then it let me login.

Finslove:
> I actually was able to get around this. I reset my password successfully. Then it let me login.

OMG, I'm such an idiot that I did not think of that myself :-D Anyway, thanks a lot @fumbles. @amruthpillai please keep up the good work. You are helping a lot of people with this tool

Overloader6:
Is there some way to find out what email I used back then?

PR #3136, opened 2026-06-01
Problem in word wrapping in the templates

Cantale08:
#fix:3107, problem with Word Breaking solved

## Bug Fixes

* Improved text hyphenation for PDFs containing mixed CJK and Latin characters. Latin words now remain unbroken while CJK content preserves character-level breaking behavior for better readability.


amruthpillai (approved):
Reviewed the CJK hyphenation fix. The branch keeps CJK character-level breaking while preserving Latin words as whole tokens in CJK-mode resumes, and the regression assertion covers the behavior. No changes needed from my side; this is merge-ready. Note that #3158 touches the same hyphenation path and should preserve this guard if both are merged.

PR #3149, opened 2026-06-06
feat: add flag to disable api rate limit

hackardoX:
### Add **FLAG_DISABLE_API_RATE_LIMIT** environment variable

Rate limiting in the auth layer is currently gated on `NODE_ENV === "production"`, which means running the app locally in production mode (e.g. a private homelab) enforces rate limits that can block legitimate API-driven workflows like managing resumes programmatically via tools such as Job Ops. Setting NODE_ENV=development is not a viable workaround since it changes other behaviors.

This PR introduces a **FLAG_DISABLE_API_RATE_LIMIT** feature flag (defaults to false) so rate limiting can be explicitly disabled while keeping NODE_ENV=production.

## Summary by CodeRabbit

* **New Features**
  * Added a new feature flag configuration option to control API rate limiting behavior for authentication endpoints. Self-hosted administrators can now optionally disable rate limiting as needed for their specific deployment scenarios. Rate limiting remains enabled by default in production environments for enhanced security.


amruthpillai (approved):
Reviewed the rate-limit feature flag change. The env schema, `.env.example`, Docker self-hosting docs, auth config usage, and `turbo.json` global env entry are all updated, which covers the strict-env gotcha for new server vars. No changes needed from my side; this is merge-ready.

PR #3158, opened 2026-06-11
fix: register language-specific Noto fallback fonts for non-Latin scripts

tkp12345:
## Screenshots
image

## Problem

Non-Latin text renders as garbled tofu boxes (□□□) or wrong glyphs in the resume builder's live preview **and** the exported PDF. The live preview is not HTML — it generates a PDF via `@react-pdf/renderer` and rasterizes it with PDF.js.

This is effectively a **regression from the v5.1.0 react-pdf migration**. Before v5.1 the preview rendered in the browser, where the OS/browser automatically substitutes any font that has the missing glyph (universal fallback). react-pdf has **no automatic system-font fallback** — a glyph only renders if a registered font contains it. So every script the chosen font lacks now breaks unless we explicitly register a fallback. CJK was patched script-by-script (#3016, #3070, #3080); Korean still mis-fell-back to a Han-only Simplified font, and Arabic/Hebrew/Thai had **no fallback at all** (the CJK detection in `cjk-regex` never matched them).

Concretely, the old fallback always returned `Noto Sans/Serif SC` (Simplified Chinese) regardless of language. `Noto Sans/Serif SC` contains CJK-Unified ideographs (Hanja/Kanji) but **not Hangul, Kana, Arabic, Hebrew or Thai**, so:
- Korean on a Latin/Chinese font → Hangul tofu (reported)
- Persian/Arabic (e.g. `Vazirmatn`, `Noto Sans Arabic`) → broken/overlapping glyphs (reported)
- Hebrew, Thai → same class of failure

## Fix

Generalize the CJK-only fallback into a **script-aware, ordered fallback stack** that covers every supported non-Latin script. react-pdf resolves the font per-codepoint left-to-right across the array, so a single resume can mix Latin with any of these:

| Script | Fallback font (serif / sans) |
|--------|------------------------------|
| Hangul (Korean) | Noto Serif KR / Noto Sans KR |
| Kana (Japanese) | Noto Serif JP / Noto Sans JP |
| Traditional Han | Noto Serif TC / Noto Sans TC |
| Simplified Han  | Noto Serif SC / Noto Sans SC |
| Arabic / Persian | Noto Naskh Arabic / Noto Sans Arabic |
| Hebrew | Noto Sans Hebrew (no Noto Serif Hebrew → reuse sans) |
| Thai | Noto Sans Thai (no Noto Serif Thai → reuse sans) |

Selection uses two signals: the resume **locale** (e.g. `ko-KR`→Hangul, `fa-IR`→Arabic, `he-IL`→Hebrew, `th-TH`→Thai — ordered first as the dominant language) and the **scripts actually detected in the content** (so a Latin-locale resume containing any of these still resolves). A Simplified-Chinese entry is appended as a safety net **only when the stack already contains a CJK script**, preserving prior `zh-CN` behavior without polluting non-CJK resumes.

CJK per-character line breaking stays **CJK-only**: it must not apply to Arabic (cursive, joined letters) or Thai (combining marks), which would break shaping.

Example stacks (serif primary `IBM Plex Serif`):
- `ko-KR` → `[IBM Plex Serif, Noto Serif KR, Noto Serif SC]`
- `fa-IR` → `[IBM Plex Serif, Noto Naskh Arabic]`
- `he-IL` → `[IBM Plex Serif, Noto Sans Hebrew]`
- `th-TH` → `[IBM Plex Serif, Noto Sans Thai]`
- `zh-CN` → `[IBM Plex Serif, Noto Serif SC]` (unchanged)

Adding a future script is now a one-line table entry (locale map + Noto font + detection regex), not a new code path.

### Changes
- `packages/utils/src/locale.ts` — generalize `CjkScript`→`Script` (+ Arabic/Hebrew/Thai), add `getLocaleScript()`, `isCjkScript()`.
- `packages/fonts/src/index.ts` — per-script Noto table + `getPdfFallbackFontFamilies()` (ordered, deduped, webfontlist-filtered, CJK-only SC safety net); keep `getPdfCjkFallbackFontFamily()` for back-compat.
- `packages/pdf/src/hooks/use-register-fonts.ts` — detect content scripts (`resumeContentScripts`), register the fallback stack, gate CJK line-breaking on CJK only.
- `packages/pdf/src/document.tsx` — pass detected scripts into `registerFonts`.

## Testing
- `@reactive-resume/fonts`: 61/61 pass.
- `@reactive-resume/pdf`: 244/244 pass (Korean/Japanese/Chinese/Arabic/Hebrew/Thai registration + script-detection + "no CJK line-break for Arabic" tests).
- `typecheck` (utils/fonts/pdf), `turbo boundaries`, and `biome check` all pass.

Manual: open the builder with a Latin font and type each script (`안녕하세요`, `سلام` / `پژوهشگر`, `שלום`, `สวัสดี`) → preview + PDF export render correctly; Chinese/Japanese still render.

### Notes / out of scope
- Full Arabic/Hebrew **bidi shaping** (contextual joining, RTL ordering) is handled by react-pdf's text engine; this PR fixes glyph **availability** (the reported tofu/Latin-fallback). Thai lacks dictionary-based word wrapping (no spaces) — glyphs render correctly but long Thai lines may not wrap; intentionally not force-broken per character (would split combining marks).
- `Vazirmatn` specifically is still absent from `webfontlist.json`; selecting it falls back to the Arabic Noto stack. Adding it to the font list is a separate change.

## New Features
- Enhanced locale-aware PDF font fallback for Korean, Japanese, Chinese (Traditional/Simplified), Thai, Arabic, and Hebrew using per-writing-system font stacks.
- Added resume-content script detection to drive smarter font pairing and ordering for multilingual resumes.

## Bug Fixes
- Prevented CJK-specific line breaking for non-CJK fallbacks (e.g., Arabic), improving typographic correctness.

## Tests
- Expanded unit coverage for fallback ordering, script detection behavior, deduplication, and web-font availability checks.


packages/pdf/src/hooks/use-register-fonts.ts
```diff
@@ -157,11 +157,72 @@ export const resumeContentContainsCJK = (data: ResumeData): boolean => {
...
+	// CJK needs per-character line breaking. This must stay CJK-only: applying
+	// it to Arabic (cursive, joined letters) or Thai (combining marks) would
+	// break shaping, so non-CJK fallbacks below do not enable it.
+	const needsCjkLineBreaking = isCJKLocale(locale) || hasCjkContent;
 
 	Font.registerHyphenationCallback((word) => {
-		if (needsCjkTextSupport) {
+		if (needsCjkLineBreaking) {
 			if (word === " ") return ["\u200C "];
 			return [...word].flatMap((l) => [l, ""]);
```
amruthpillai: This brings back the bug fixed in #3136: once `needsCjkLineBreaking` is true, Latin words in a CJK resume are split into individual letters. Gate the per-character split on the word actually containing CJK characters, and keep/add the `hyphenationCallback?.("Reactive")` regression test from #3136.

```suggestion
			if (cjkLetterRegex.test(word)) {
				return [...word].flatMap((l) => [l, ""]);
			}
```

amruthpillai (changes requested):
I found one blocking regression in the hyphenation path. The script-aware fallback work looks directionally right, but this branch currently reintroduces the Latin word-splitting behavior that #3136 fixes: when CJK line breaking is enabled, every word is split character-by-character, including `Reactive` and other Latin tokens in CJK resumes. Please keep the fallback stack changes, but retain the per-word CJK guard and its test before merging.

amruthpillai (approved):
Re-reviewed the updated branch at `3923f604`. The previous blocker is fixed: CJK per-character line breaking is now gated by `cjkLetterRegex.test(word)`, and the regression test keeps Latin words like `Reactive` intact in CJK mode.

Focused validation passed locally:
- `pnpm --filter @reactive-resume/pdf test -- src/hooks/use-register-fonts.test.ts`
- `pnpm --filter @reactive-resume/fonts test -- src/index.test.ts`
- `pnpm --filter @reactive-resume/pdf typecheck`
- `pnpm --filter @reactive-resume/fonts typecheck`
- `pnpm --filter @reactive-resume/utils typecheck`
- `pnpm exec turbo boundaries`

No remaining blockers from my side. Good to merge.

PR #3163, opened 2026-06-17
fix(ci): make Docker Hub publishing optional

GoldenSection0618:
## Summary

- Make Docker Hub publishing optional in the Docker image workflow.
- Publish GHCR images even when Docker Hub credentials are not configured.
- Normalize image names to lowercase so fork owner names with uppercase characters still produce valid image references.
- Skip release-only SSH redeploy and Cloudflare purge steps when their secrets are not configured.

## Why

Forks can run the Docker image workflow without the upstream repository secrets. The workflow currently assumes Docker Hub credentials are always present, so it fails during Docker Hub login with `Username and password required`. After that is fixed, fork-triggered release-style runs can also fail on missing deployment secrets.

This keeps the upstream behavior unchanged when the secrets exist, while allowing forks to build, publish, sign, and inspect GHCR images successfully without Docker Hub or deployment credentials.

## Validation

- `git diff --check`
- Parsed `.github/workflows/docker-build.yml` with `pnpm dlx js-yaml`
- `bash -n` on the changed workflow shell blocks
- Successful workflow run on the fork: https://github.com/GoldenSection0618/reactive-resume/actions/runs/27692773706

### DIFF
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -66,7 +66,32 @@ jobs:
       - name: Setup Docker Buildx
         uses: docker/setup-buildx-action@v4
 
+      - name: Determine registries
+        id: registries
+        env:
+          DOCKER_USERNAME: ${{ secrets.DOCKER_USERNAME }}
+          DOCKER_PASSWORD: ${{ secrets.DOCKER_PASSWORD }}
+        run: |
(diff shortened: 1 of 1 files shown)


amruthpillai:
Good idea, looking forward to the complete PR

PR #3165, opened 2026-06-18
fix(fonts): register Noto Armenian fallback fonts for Armenian script

agasy18:
## Problem

Armenian text (e.g. a resume written in Հայերեն) renders as tofu (□□□) in **both the PDF export and the on-screen preview**. react-pdf only renders glyphs from registered fonts, and for a resume using a Latin font (e.g. the default IBM Plex Serif) no Armenian-capable font is registered, so Armenian codepoints have no glyph.

## Root cause

v4 bundled the full Google Fonts catalog, which included `Noto Sans Armenian` and `Noto Serif Armenian`. v5 trimmed this to a curated `webfontlist.json` and introduced a per-script Noto fallback mechanism — but that mechanism only covers CJK, Arabic, Hebrew and Thai. Armenian was dropped from the catalog and never added to the fallback system, so it regressed.

## Fix

Adds Armenian to the existing per-script fallback mechanism, mirroring the Arabic/Hebrew/Thai pattern exactly:

- **`packages/fonts/src/webfontlist.json`** — add `Noto Sans Armenian` + `Noto Serif Armenian` (all weights).
- **`packages/utils/src/locale.ts`** — add `"armenian"` to the `Script` union.
- **`packages/fonts/src/index.ts`** — map `armenian → { serif: "Noto Serif Armenian", sansSerif: "Noto Sans Armenian" }`. Unlike Hebrew/Thai, a real Noto **Serif** Armenian exists, so the serif slot uses it rather than reusing the sans font.
- **`packages/pdf/src/hooks/use-register-fonts.ts`** — detect the Armenian block (U+0531–U+058F) and Armenian ligatures (U+FB13–U+FB17) in resume content so the fallback is auto-registered.

Armenian content now resolves against a Noto Armenian font in both the preview and the PDF regardless of the selected primary font, and the fonts remain directly selectable. The detector does not match Latin or Cyrillic, so other languages are unaffected.

## Testing

- `@reactive-resume/fonts` — 62/62 pass (incl. new Armenian fallback-selection cases)
- `@reactive-resume/pdf` — 246/246 pass (incl. new Armenian detection + registration cases)
- `@reactive-resume/utils` `locale.test.ts` — 22/22 pass
- `biome check` on all changed files — clean

## Notes

- No schema/API/DB/migration surface is touched.
- One small judgment call worth flagging: the serif slot maps to the real `Noto Serif Armenian` (a real serif variant exists), whereas Hebrew/Thai reuse their sans font for serif because no serif Noto exists for those scripts. Happy to align to the reuse pattern if maintainers prefer consistency over correct serif glyphs.

## Summary by CodeRabbit

* **New Features**
  * Added Armenian language support with native fonts in PDF generation, including automatic script detection for Armenian text

* **Tests**
  * Added test coverage for Armenian script detection and font fallback behavior

### DIFF
--- a/packages/fonts/src/index.test.ts
+++ b/packages/fonts/src/index.test.ts
@@ -196,6 +196,11 @@ describe("getPdfFallbackFontFamilies", () => {
 		expect(getPdfFallbackFontFamilies("Helvetica", { locale: "th-TH" })).toEqual(["Noto Sans Thai"]);
 	});
 
+	it("uses the category-matching Armenian Noto font (real sans and serif variants exist)", () => {
+		expect(getPdfFallbackFontFamilies("Helvetica", { scripts: ["armenian"] })).toEqual(["Noto Sans Armenian"]);
+		expect(getPdfFallbackFontFamilies("Times-Roman", { scripts: ["armenian"] })).toEqual(["Noto Serif Armenian"]);
+	});
+
 	it("does not append the Simplified Chinese safety net for non-CJK scripts", () => {
 		expect(getPdfFallbackFontFamilies("Helvetica", { scripts: ["arabic"] })).toEqual(["Noto Sans Arabic"]);
 		expect(getPdfFallbackFontFamilies("Helvetica", { scripts: ["thai"] })).not.toContain("Noto Sans SC");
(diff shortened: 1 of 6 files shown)


amruthpillai:
@agasy18 are you still open to working on this? There seems to be a valid review comment that needs addressing before this can be merged.

Issue #3185, opened 2026-06-29
[Bug] Right edge of text cutt of or clipped in exported PDF and also on web preview (Cloud)


nkohade24:
### Is there an existing issue for this?

- [x] Yes, I have searched the existing issues and none of them match my problem.

### Product Variant

Cloud (https://rxresu.me)

### Describe the bug you're experiencing

In the Experience section, text on the right side of the page is cut off when it appears in bulleted lines. The last letters of words at the right edge are missing. This happens in the web preview and also in the exported PDF.

<img width="907" height="533" alt="Image" src="https://github.com/user-attachments/assets/c5c0bbe7-6df2-454c-908a-20a16ed3bddf" />

Without bullets they work fine 

<img width="882" height="525" alt="Image" src="https://github.com/user-attachments/assets/d0436678-5ebd-4448-ae42-4423bb12bed3" />

expected result: text should wrap within the page margins and no characters should be cut off.

### What template are you using?

Onyx

jshn9515:
And for Chinese with Rhyhorn template

<img width="972" height="293" alt="Image" src="https://github.com/user-attachments/assets/c21fc7d9-5795-4a80-b849-9396b0d8c715" />

SimoHypers:
found the root cause - overflow: hidden in safeTextStyle was being applied to all Text primitives in the PDF renderer. in @react-pdf/renderer v4.x this clips the element at its initial computed height, so any text after a line break just gets hidden.

fix is a one-liner: remove overflow: hidden from safeTextStyle. the other properties (minWidth: 0, maxWidth: 100%, flexShrink: 1) already handle horizontal containment so nothing else breaks. opening a PR with the fix now.

Issue #3184, opened 2026-06-29
[Bug] <title>Text not wrapping


JP0047:
### Is there an existing issue for this?

- [x] Yes, I have searched the existing issues and none of them match my problem.

### Product Variant

Cloud (https://rxresu.me)

### Describe the bug you're experiencing

The text in the experience section and below do not wrap appropriately as some words get cut off at the margin  

### What template are you using?

Ditgar

JP0047:
<img width="1256" height="771" alt="Image" src="https://github.com/user-attachments/assets/9a1c8762-27b3-4ade-a81c-ba73c0bd9da7" />

PR #3186, opened 2026-06-29
fix: remove overflow hidden from safeTextStyle

SimoHypers:
found the root cause of the text wrapping issue - `overflow: hidden` in `safeTextStyle` was being applied to all Text primitives in the PDF renderer. in `@react-pdf/renderer` v4.x this clips the element at its initial computed height, so any text after a line break just gets hidden.

fix is a one-liner: remove `overflow: hidden` from `safeTextStyle`. the other properties (`minWidth: 0`, `maxWidth: 100%`, `flexShrink: 1`) already handle horizontal containment so nothing else breaks.

also removed the stale comment in `styles.ts` that was documenting a lineHeight workaround for the same overflow clipping issue on headings.

## changes
- `packages/pdf/src/templates/shared/safe-text-style.ts` - removed `overflow: hidden`
- `packages/pdf/src/templates/shared/styles.ts` - removed stale workaround comment

## Summary by CodeRabbit

* **New Features**
  * Added improved link-style handling for PDF content, helping link appearance stay consistent when styling is combined.

* **Bug Fixes**
  * Adjusted text styling to avoid clipping issues in narrow layouts by changing how overflow is handled.
  * Cleaned up heading spacing behavior without changing the visible layout settings.

referencing issue #3184 and #3185

### DIFF
--- a/packages/pdf/src/templates/shared/safe-text-style.ts
+++ b/packages/pdf/src/templates/shared/safe-text-style.ts
@@ -4,5 +4,4 @@ export const safeTextStyle = {
 	minWidth: 0,
 	maxWidth: "100%",
 	flexShrink: 1,
-	overflow: "hidden",
 } satisfies Style;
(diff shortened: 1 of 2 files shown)


builtbyshail:
Just tried this out on my local and can confirm that it works!

amruthpillai:
Thank you for the PR!

ASHMITA-DE:
Please merge the changes

Issue #3201, opened 2026-07-04
[Bug] Bullet list items overflow past page margin in export and preview


shanuflash:
### Is there an existing issue for this?

- [x] Yes, I have searched the existing issues and none of them match my problem.

### Product Variant

Cloud (https://rxresu.me)

### Describe the bug you're experiencing

Bullet points in rich-text fields (e.g. Experience, Summary) render past the right page margin in the exported PDF instead of wrapping. Long bullet lines run off the edge of the page.

### What template are you using?

None

### Anything else?

In the PDF list-item renderer (`packages/pdf/src/templates/shared/rich-text.tsx`), the `<li>` content is placed in a row (`flexDirection: "row"`) alongside the bullet marker, but the content is only given `flex: 1` when the item contains a nested list. For plain-text bullets the content has no flex constraint, so the text takes its intrinsic width and never wraps.

BeckKeller:
@shanuflash and @hackardoX, I'm still experiencing this issue post-PR merge across multiple resume templates (Rhyhorn, Lapras, etc), and in both new and existing resumes. Please let me know if I should open a separate bug report.

hackardoX:
Hey @BeckKeller, did you test locally with the latest main? Or on the cloud? Because it has not been released yet

BeckKeller:
@hackardoX Ah, that's why. I'm testing in the cloud. Thank you!

For those also experiencing this issue, inserting a manual line break serves as a workaround.

Issue #3177, opened 2026-06-23
[Bug] Bullet point marker orphaned on previous page when <li> breaks across pages


hackardoX:
### Is there an existing issue for this?

- [x] Yes, I have searched the existing issues and none of them match my problem.

### Product Variant

Self-Hosted

### Describe the bug you're experiencing

When a bullet list item in the experience section is close to a page break, the content wraps to the next page but the bullet marker (•) stays on the previous page.

### Root cause:
The \<li> renderer in packages/pdf/src/templates/shared/rich-text.tsx placed the marker (<PdfText>•</PdfText>) and the content as separate children of a flexDirection: "row" <View>. react-pdf's page breaker can split flex children, orphaning the marker on the previous page.

### Fix: 
Wrapped the marker and the first content child in a nested `<View style={{ flexDirection: "row" }}>` inside the LTR content node, ensuring they stay together as a single unit during page breaks. Remaining children (rare in practice) can still break independently.

### Before:

<img width="509" height="251" alt="Image" src="https://github.com/user-attachments/assets/fa408c75-67c9-4db6-b3be-4e0f067b3e1c" />

### After: 

<img width="510" height="275" alt="Image" src="https://github.com/user-attachments/assets/cb635513-0123-4a29-805d-d032eb83acaa" />

### What template are you using?

Kakuna

PR #3236, opened 2026-07-08
fix: solve list marker page break (#3177)

hackardoX:
The fix introduces to solve the marker orphan seemed to have introduced regression (#3201 and #3202). I decided to revert the state before #3177, and tackle the first issue accordingly. So, the change consists in:
- Add flex: initial to list row element (marker and text content)
- Add minPresenceAhead to wrap the marker in case not enough space at the end of the page

Before:
Screenshot 2026-07-08 at 23 27 05

After:

After

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved list item layout in PDF-rich text so markers and content align more consistently in both left-to-right and right-to-left text.
  * Adjusted spacing behavior to prevent list content from stretching unexpectedly, resulting in cleaner and more predictable list formatting.


hackardoX:
Hi @amruthpillai, can you review this PR whenever you have some time? Thanks

PR #3319, opened 2026-08-15
fix(ai): bound the provider test and explain why it failed

### LINKED ISSUE #3073: [bug] AI provider Test Connection behaves like full generation validation and can appear stuck

### Product Variant

Cloud (https://rxresu.me)

The AI provider **Test Connection** action is misleading and can appear stuck.

The UI suggests a lightweight connectivity/API-key check, but the implementation performs a real `generateText()` request with `Output.choice()`, requiring the selected model to return a structured result.

Because of this, the test can take much longer than expected, especially with slower gateways, local models, or OpenAI-compatible providers. There is also no visible timeout or explanation that a full model generation request is being performed.

This makes it hard to tell whether the delay is caused by network connectivity, invalid credentials, provider latency, model behavior, or structured output compatibility.

I think this should either be presented as a model generation test, or made closer to a lightweight connection test.

None

### Anything else?

_No response_

MrTig-afk:
Closes #3073.

## Background

The issue is partly stale — the `Output.choice()` complaint was fixed by a previous change, and `testConnection` is now a plain `generateText`. Two problems remain.

**1. The test could hang.** `generateText` is called with no `abortSignal` and no `timeout`, and the AI SDK defaults to `maxRetries: 2`, so a failing call is attempted three times. Nothing in the stack bounds it, which is the "appears stuck" report.

**2. Every failure looks the same.** `testConnection` returns a bare `boolean`. The raw SDK error is persisted to `testError`, but the router flattens anything non-`ORPCError` to a fixed `"Could not reach the AI provider."`, so a rejected key, a wrong model name and an unreachable host are indistinguishable.

## What the user sees now

| Situation | Before | After |
|---|---|---|
| Wrong API key | `Incorrect API key provided.` | `OpenAI rejected the API key.` |
| Wrong model name | *(empty string)* | `OpenAI has no model named "gpt-fake-9", or the base URL is wrong.` |
| Provider is down | `Failed after 3 attempts. Last error: AI_APICallError` | `Groq reported a server error (503). This is a problem on the provider's side.` |
| Base URL unreachable | `Failed after 3 attempts. Last error: AI_APICallError: Cannot connect to API: connect ECONNREFUSED 127.0.0.1:11434` | `Could not reach http://127.0.0.1:11434/api. Check that the base URL is correct and the service is running.` |
| Provider hangs | never returns | `Ollama did not respond within 30 seconds. The service may be unreachable, or the model may be too slow to load.` |
| Model ignores the prompt | `The provider test returned an unexpected response.` | `The provider is reachable, but the model replied with unexpected output instead of a simple confirmation.` |

## Changes

- **Bounded** (`ai/service.ts`): a 30s `AbortSignal.timeout` plus `maxRetries: 0`, so the test cannot silently triple its own wait.
- **Classified** (`ai/service.ts`): failures are mapped from the SDK's own types — `APICallError` status codes, `LoadAPIKeyError`, `NoSuchModelError`, socket codes found by walking the `cause` chain, and aborts — into one actionable sentence. The API key is scrubbed from any message passed through.
- **Returned as data** (`ai-providers/service.ts`): a provider-side failure is a completed test, not a failed request, so it comes back with `testStatus: "failure"` and the reason. The client already had branches for this in both the provider card and the add-provider form, but the server never sent it, so they were unreachable. Infrastructure errors still throw.
- **Names** (`packages/ai/src/types.ts`): `AI_PROVIDER_DISPLAY_NAMES`, beside the existing base-URL map, so messages can name the provider.
- **Wait cue** (`ai-section.tsx`): the spinner picks up an elapsed counter after 5s and names the provider after 20s. The toast is reduced to the outcome, since the card already shows the reason.

`getModel()` is deliberately resolved outside the try block: it runs the base-URL policy check, and that error must keep reaching the router as `BAD_REQUEST` rather than being reported as a provider failure. There is a test pinning this.

## Verification

`packages/api/src/features/ai-providers/e2e.test.ts` drives the real router (auth middleware and error mapping included), service, `testConnection`, classifier and AES-GCM credential encryption, substituting only the database and the outbound provider call.

Reverting just the source files to `main` and re-running: **5 of the 8 cases fail**, covering each behaviour change. The other 3 are invariants that pass in both — a healthy provider still enables, a blocked base URL still raises `BAD_REQUEST`, and an unknown id still raises `NOT_FOUND`. The no-retry case is also visible in wall-clock: 5007 ms before, instant after.

Also run: `pnpm --filter @reactive-resume/api test`, `pnpm --filter web test`, `turbo boundaries`, `knip`, Biome, and `tsgo --noEmit` on the touched packages.

## Note for reviewers

The classified messages are composed server-side and therefore English-only, consistent with the existing messages in `router.ts`. Making them translatable would mean returning reason codes and rendering them client-side, which needs a new column and a migration — happy to do that in a follow-up if you'd prefer it.

## Summary by CodeRabbit

* **New Features**
  * AI provider connection tests now show elapsed time and a “Still waiting” status for longer checks.
  * Added human-readable names for supported AI providers.
  * Added localized connection-testing status messages.

* **Bug Fixes**
  * Improved failure handling with clear, redacted provider error details.
  * Disabled retries and added a 30-second timeout for connection checks.

* **Tests**
  * Expanded coverage for credentials, outages, timeouts, malformed responses, and sensitive-data redaction.

The PR bounds AI provider connection tests and returns classified, credential-redacted failure reasons to the client.
- Disables automatic retries and applies a 30-second abort signal.
- Persists provider test outcomes as structured success or failure data.
- Adds provider display names, elapsed waiting feedback, and end-to-end coverage.
</details>

The PR appears safe to merge.

No blocking failure remains.
</details>

| Filename | Overview |
|----------|----------|
| packages/api/src/features/ai/service.ts | Bounds connection generation, classifies provider and transport failures, and redacts credentials from forwarded messages. |
| packages/api/src/features/ai-providers/service.ts | Persists completed provider failures as response data while retaining thrown handling for unexpected failures. |
| packages/api/src/features/ai-providers/e2e.test.ts | Exercises the router-to-provider-test flow across success, classification, timeout, retry, redaction, and policy cases. |
| apps/web/src/features/settings/integrations/components/ai-section.tsx | Adds elapsed waiting labels and presents concise connection-test outcomes while retaining detailed card errors. |
| packages/ai/src/types.ts | Adds exhaustive display names for every supported AI provider. |

</details>

<details open><summary><h3>Flowchart</h3></summary>

```mermaid
%%{init: {'theme': 'neutral'}}%%
flowchart TD
    A[User tests AI provider] --> B[Resolve model and base URL policy]
    B -->|Invalid configuration| C[Router returns BAD_REQUEST]
    B -->|Valid| D[generateText with no retries and 30s abort]
    D -->|Confirmation received| E[Persist success and enable provider]
    D -->|Provider or network failure| F[Classify and redact reason]
    F --> G[Persist failure and return reason as data]
    D -->|Unexpected infrastructure failure| H[Persist failure and rethrow]
```
</details>

<sub>Reviews (4): Last reviewed commit: ["chore(i18n): extract catalogs for the pr..."](https://github.com/amruthpillai/reactive-resume/commit/9251ca54eb9708b186a576b2d08d6e402fed444c) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=53571820)</sub>

### DIFF
--- a/apps/web/locales/af-ZA.po
+++ b/apps/web/locales/af-ZA.po
@@ -986,6 +986,10 @@ msgid "Connection failed"
 msgstr "Verbinding het misluk"
 
 #: src/features/settings/integrations/components/ai-section.tsx
+msgid "Connection failed."
+msgstr ""
+
+#: src/features/settings/integrations/components/ai-section.tsx
 msgid "Connection verified — provider is ready to use."
 msgstr "Verbinding geverifieer — verskaffer is gereed vir gebruik."
 
@@ -4309,6 +4313,10 @@ msgstr "Statistiek"
 msgid "Stay"
 msgstr "Bly"
 
+#: src/features/settings/integrations/components/ai-section.tsx
+msgid "Still waiting for {provider}… {elapsedSeconds}s"
+msgstr ""
+
 #: src/routes/agent/-components/agent-chat.tsx
 msgid "Stop generation"
 msgstr "Stop generasie"
@@ -4440,6 +4448,10 @@ msgstr "Getuigskrifte"
 #: src/features/settings/integrations/components/ai-section.tsx
 msgid "Testing…"
 msgstr "Toetsing…"
+
+#: src/features/settings/integrations/components/ai-section.tsx
+msgid "Testing… {elapsedSeconds}s"
+msgstr ""
 
 #: src/components/input/rich-input.tsx
 msgid "Text alignment"
@@ -5399,4 +5411,3 @@ msgstr "Zoem uit"
 #: src/libs/locale.ts
 msgid "Zulu"
 msgstr "Zoeloe"
-
--- a/apps/web/locales/am-ET.po
+++ b/apps/web/locales/am-ET.po
@@ -986,6 +986,10 @@ msgid "Connection failed"
 msgstr "ግንኙነት አልተሳካም"
 
 #: src/features/settings/integrations/components/ai-section.tsx
+msgid "Connection failed."
+msgstr ""
+
+#: src/features/settings/integrations/components/ai-section.tsx
 msgid "Connection verified — provider is ready to use."
 msgstr "ግንኙነት ተረጋግጧል - አቅራቢው ለመጠቀም ዝግጁ ነው።"
 
@@ -4309,6 +4313,10 @@ msgstr "ስታቲስቲክስ"
 msgid "Stay"
 msgstr "ቆይ"
 
+#: src/features/settings/integrations/components/ai-section.tsx
+msgid "Still waiting for {provider}… {elapsedSeconds}s"
+msgstr ""
+
 #: src/routes/agent/-components/agent-chat.tsx
 msgid "Stop generation"
 msgstr "ማመንጨት አቁም"
@@ -4440,6 +4448,10 @@ msgstr "ምስክርነቶች"
 #: src/features/settings/integrations/components/ai-section.tsx
 msgid "Testing…"
 msgstr "ሙከራ…"
+
+#: src/features/settings/integrations/components/ai-section.tsx
+msgid "Testing… {elapsedSeconds}s"
+msgstr ""
 
 #: src/components/input/rich-input.tsx
 msgid "Text alignment"
@@ -5399,4 +5411,3 @@ msgstr "አጉር"
 #: src/libs/locale.ts
 msgid "Zulu"
 msgstr "ዙሉ"
-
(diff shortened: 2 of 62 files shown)


MrTig-afk:
Good catch — fixed in 9756ad5.

`aiCredentialsSchema` does accept `min(1)`, so a short key was reachable, and a leaked credential is worse than a mangled message. Removed the threshold and kept the empty-string guard, which is only there to stop `replaceAll` splicing the placeholder between every character.

Added a regression test alongside the existing one, using a six-character key so it fails against the previous threshold:

```ts
it("redacts a short API key as well", async () => {
	stubFailedResponse(400, JSON.stringify({ error: { message: "Bad key tok123" } }));

	const result = await testConnection(testInput({ apiKey: "tok123" }));

	expect(result.ok).toBe(false);
	if (!result.ok) {
		expect(result.message).not.toContain("tok123");
		expect(result.message).toContain("***");
	}
});
```

PR #3387, opened 2026-08-27
fix(components/form): resolve FormControl label target regressions (#3369)

### LINKED ISSUE #3369: FormControl stamps the label's target id on a non-labelable wrapper, so Slug, Tags and the Sidebar Width slider have no accessible name
### Existing issue

- [x] I searched the existing issues and could not find a matching report.

Keyword searches for screen readers, aria labels and accessible names, plus a sweep of the 800 most recent issues and pull requests, found one open accessibility item and one closed precedent. The open one, https://github.com/AmruthPillai/Reactive-Resume/issues/2844, is about the **rendered resume output** (heading levels, semantic bullets, image alt text), a different surface from the builder's form controls, so this is not a duplicate of it. It is worth noting that its only maintainer comment asks the reporter to narrow the report to a concrete accessibility problem with current behaviour, expected behaviour and acceptance criteria, which is the shape this report already has. The closed https://github.com/AmruthPillai/Reactive-Resume/issues/2298 ("buttons do not have accessible names", v4 marketing homepage) is precedent that reports of this class are accepted here.

### Product variant

Self-hosted

### Reactive Resume version

5.2.7 (commit `3221afda9ddfb03d6cce87927b0ce47338b4cfa8` on `main`, 16 commits after the `v5.2.7` tag, so this exact build is not a release)

### Area

Resume builder & data

### Environment

Chromium 141.0.7390.37 (headless, driven by `playwright-core` 1.62.1, using Playwright's own accessibility computation) on macOS (Darwin 25.5.0, arm64); self-hosted, built from source with `pnpm install --frozen-lockfile` and `pnpm run build`, run as `node apps/server/dist/index.mjs`, PostgreSQL 18 in Docker.

### Summary

`FormLabel` renders an `htmlFor` of the form `<id>-form-item`, and `FormControl` puts that same id on whatever element the render prop's root happens to be. Where the render prop is a **wrapper** rather than the real control, the id lands on an element that a `<label>` cannot name, and the real input inside it is left anonymous.

Measured live, resolving each `label[for]` to the element that actually carries the id:

```
label "Name"          for=_r_3h_-form-item  -> target=INPUT             OK
label "Slug"          for=_r_3j_-form-item  -> target=FIELDSET          BROKEN
label "Tags"          for=_r_3l_-form-item  -> target=DIV               BROKEN
label "Sidebar Width" for=_r_jv_-form-item  -> target=DIV[role=group]   BROKEN
```

A `<label>` can only name a labelable element, so for the three broken rows the visible caption is not attached to anything: the text input inside the Slug fieldset, the chip field behind Tags, and the slider thumb behind Sidebar Width are all anonymous to assistive technology and to any name-based query. In the same dialog `getByLabel('Slug')` and `getByLabel('Tags')` both count **0**, while `getByLabel('Name')` counts 1.

Two corrections to any earlier or broader version of this claim, which we make up front so a maintainer opening the builder does not find them and discount the whole report:

- **Exactly 2 of the 15 builder spinbuttons are anonymous, not 15.** The other 13 are correctly named. The two are Picture -> Size (value `"100"`) and the Sidebar Width numeric input (value `"30"`).
- The pattern is real but not everywhere. Several places that look like instances are correctly nested (`InputGroup` wrapping `FormControl render={<InputGroupInput …>}`) and measure as correctly named.

Our run also reproduced a **second defect family** in the same components, which is smaller but concrete: `FormLabel`s whose `for=` names an id that no element in the document carries at all (Picture -> Size, Basics -> Website), and one duplicate DOM id (`_r_jv_-form-item` appears twice) because one `FormItem` contains two `FormControl`s.

### Steps to reproduce

**A. The Create-a-resume dialog.**

1. Sign in and open the dashboard, then press the control that opens **Create a new resume**.
2. Inspect the accessibility tree, or run the queries below. `Name` resolves; `Slug` and `Tags` do not.

Verbatim from our run, the relevant part of the dialog's accessibility snapshot:

```
- dialog "Create a new resume":
  - text: Name
  - textbox "Name"
  - button "Generate a random name"
  - text: Slug
  - group:
    - group: http://127.0.0.1:54490/guardowner/
    - textbox                       <- anonymous
  - text: Tags
  - textbox "Add keyword"
  - group "Create resume with options":
    - button "Create"
```

```
getByLabel('Name').count()                               1
getByLabel('Slug').count()                               0
getByLabel('Tags').count()                               0
CONTROL getByRole('textbox', {name:'Name', exact:true})  1
getByRole('textbox', {name:'Slug'})                      0

label "Name"  for=_r_3h_-form-item  -> target=INPUT      OK
label "Slug"  for=_r_3j_-form-item  -> target=FIELDSET   BROKEN
label "Tags"  for=_r_3l_-form-item  -> target=DIV        BROKEN
```

**B. The builder sidebar, Layout and Typography sections open.**

3. Open a resume in the builder and expand the right sidebar's Layout, Page and Typography sections.
4. Query the Sidebar Width slider by its visible label. It does not resolve, although exactly one slider exists on the page:

```
getByRole('slider', {name: 'Sidebar Width'})   0
sliders on the page                            1
slider attributes  [{"tag":"INPUT","ariaLabel":null,"ariaLabelledby":null,"id":"base-ui-_r_k3_"}]
```

5. Take the spinbutton census. This is where the corrected count comes from:

```
spinbutton census (name / value):
  spinbutton: (ANONYMOUS): "100"          <- Picture > Size
  spinbutton "Rotation": "0"
  spinbutton "Aspect Ratio": "1"
  spinbutton "Border Radius": "0"
  spinbutton "Border Width": "0"
  spinbutton "Shadow Width": "0"
  spinbutton: (ANONYMOUS): "30"           <- Sidebar Width numeric
  spinbutton "Font Size": "10"
  spinbutton "Line Height": "1.5"
  spinbutton "Font Size": "12"
  spinbutton "Line Height": "1.5"
  spinbutton "Margin (Horizontal)": "16"
  spinbutton "Margin (Vertical)": "16"
  spinbutton "Spacing (Horizontal)": "12"
  spinbutton "Spacing (Vertical)": "8"

total spinbuttons                                    15
spinbuttons with NO accessible name                   2
getByRole('spinbutton', {name: 'Font Size'})          2   <- CONTROL: correctly named
getByRole('spinbutton', {name: 'Size', exact: true})  0
```

**C. The second family, dangling `for=` targets and a duplicate id.**

6. Resolve every `FormLabel`'s `for=` to `document.getElementById`. Two name ids that nothing in the document carries, and one `FormControl` id is used twice:

```
label "Size"          for=_r_1v_-form-item  -> target=NOTHING          (Picture > Size)
label "Website"       for=_r_2v_-form-item  -> target=NOTHING          (Basics > Website)
label "Sidebar Width" for=_r_jv_-form-item  -> target=DIV[role=group]
label "Font Family"   for=_r_kf_-form-item  -> target=BUTTON[role=combobox]

DOM ids used more than once  [["_r_jv_-form-item", 2]]
```

#### The control, which shows the instrument is fine

On the same two surfaces, correctly wired fields resolve normally: `getByLabel('Name')` = **1** in the dialog, `getByRole('textbox', {name: 'Name', exact: true})` = **1**, and `getByRole('spinbutton', {name: 'Font Size'})` = **2** in the sidebar. So the pages render, the labels exist, and the query method works. It is the specific fields listed above that have no name.

### Expected behavior

Every labelled control should be reachable by its visible label: `getByLabel('Slug')`, `getByLabel('Tags')` and `getByRole('slider', {name: 'Sidebar Width'})` should each resolve to the real input, exactly as `getByLabel('Name')` already does, and a screen reader should announce the field's caption when focus lands on it.

The fix is in the shared component rather than in each call site. `FormControl` should put the `<id>-form-item` id on the labelable element rather than on the render prop's root. Options, in the order we would try them:

1. Have `FormControl` forward the id to the inner control that the wrapper renders (for `InputGroup`, that is `InputGroupInput`; for `Slider`, the `role="slider"` element; for `ChipInput`, its inner text input), so the existing call sites keep working unchanged.
2. Or, where a wrapper genuinely has to carry the association, wire it with `aria-labelledby` pointing at the label's own id instead of `htmlFor`, which does work on a `<fieldset>` or a `div[role=group]`.
3. Separately, guard against the two smaller defects: a `FormLabel` whose `for=` resolves to nothing at all, and two `FormControl`s sharing one `FormItem` and therefore one id.

A development-time assertion would keep this from returning: on mount, `FormControl` can check that the element receiving the id is labelable, and warn when it is not. That turns a silent accessibility regression into something visible in the console the day it is written.

### Actual behavior

The id lands on the wrapper. `Slug` resolves to a `FIELDSET`, `Tags` to a `DIV`, `Sidebar Width` to a `DIV[role=group]`, and the real control inside each has no accessible name. The Sidebar Width slider is an `INPUT` with `aria-label: null` and `aria-labelledby: null`. Two of the builder's fifteen spinbuttons are anonymous. Two `FormLabel`s point at ids no element carries, and one id is present twice in the document.

#### Cause

Read at `3221afda9ddfb03d6cce87927b0ce47338b4cfa8`:

```
packages/ui/src/components/form.tsx:30-42     FormLabel -> <Label htmlFor={`${id}-form-item`} …>   (htmlFor at :38)
packages/ui/src/components/form.tsx:44-61     FormControl -> useRender({ props: { id: `${id}-form-item` } })   (id at :52)
```

https://github.com/AmruthPillai/Reactive-Resume/blob/3221afda9ddfb03d6cce87927b0ce47338b4cfa8/packages/ui/src/components/form.tsx#L30-L61

The wrappers the id lands on, and why a label cannot name them:

```
packages/ui/src/components/input-group.tsx:9-19       InputGroup renders a <fieldset>
packages/ui/src/components/input-group.tsx:117-128    InputGroupInput renders the real <Input>, and is NOT what the id lands on
apps/web/src/components/input/chip-input.tsx:330-331  ChipInput's root is a <div>
packages/ui/src/components/slider.tsx:15-25           Slider's root is SliderPrimitive.Root, not the role="slider" thumb
```

The call sites that produce the three named symptoms:

```
apps/web/src/dialogs/resume/index.tsx:393-395   FormControl render={<InputGroup>}    -> Slug
apps/web/src/dialogs/resume/index.tsx:425-433   FormControl render={<ChipInput …>}   -> Tags
apps/web/src/routes/builder/$resumeId/-sidebar/right/sections/layout/index.tsx:71-84   FormControl render={<Slider …>}
apps/web/src/routes/builder/$resumeId/-sidebar/right/sections/layout/index.tsx:86-88   a SECOND FormControl in the same FormItem  -> the duplicate id
```

A note on the variant dropdown, since it takes only one value: we drove a **self-hosted** build. This is pure client-side rendering with no deployment branch, so the cloud deployment renders the identical markup. Please do not read the dropdown as narrowing this to self-hosted installs.

### Logs and screenshots

No screenshot is needed, and a screenshot would in fact hide the bug: these fields look correctly labelled on screen. The evidence is the label-to-target resolution and the role queries quoted above, both taken from the live document.

The single line that states the whole finding:

```
label "Name" for=_r_3h_-form-item -> target=INPUT      OK
label "Slug" for=_r_3j_-form-item -> target=FIELDSET   BROKEN
label "Tags" for=_r_3l_-form-item -> target=DIV        BROKEN
```

(The `_r_…_` ids are React-generated and will differ per render; the target element types are what matter and are stable.)

#### Suggested labels

`bug`, `status: needs triage` (both applied by the form), plus `v5` and `area: builder`. Our account cannot apply labels itself.

Deliberately **no** deployment label: this is client-side rendering with no deployment branch, so narrowing it to either cloud or self-hosted would be wrong.

Found by TrueCourse running the product's published documentation against a live instance; the full transcript (the accessibility snapshots, every label-to-target resolution and the complete spinbutton census) is available on request.

santhiprakash:
fix(components/form): resolve FormControl label target regressions (#3369)

## Problem

`FormLabel` uses `htmlFor` to point at `${id}-form-item`, and `FormControl` stamps that id on whatever element `useRender` resolves to. For three call sites in the builder, the rendered root is a wrapper, so the label names nothing and the real control is anonymous:

- **Slug** (`apps/web/src/dialogs/resume/index.tsx`) — `InputGroup` renders a `<fieldset>`, while the real input is `InputGroupInput`.
- **Tags** (`apps/web/src/components/input/chip-input.tsx`) — `ChipInput` renders a `<div>`, while the inner text input carried `aria-label="Add keyword"` that overrode any wired label.
- **Sidebar Width** (`apps/web/src/routes/builder/$resumeId/-sidebar/right/sections/layout/index.tsx`) — `Slider` renders a `div[role="group"]`, while the thumb `input[type="range"]` had no accessible name.

## Fix

- Introduce `FormControlContext` so `FormControl` can pass the generated `controlId` and `labelId` to descendant controls.
- Wrap `FormControl` children in Base UI's `LabelableProvider` so the Base UI `Slider` thumb receives the correct id and `aria-labelledby`.
- Update `InputGroup` / `InputGroupInput` (and `InputGroupTextarea`) to consume the context and place the id on the real labelable element, stripping wrapper `id` / `aria-describedby` / `aria-invalid` from the `<fieldset>`.
- Update `ChipInput` to consume the context, set the inner input id, and drop the generic `aria-label` when a visible `FormLabel` is present.
- Restructure the sidebar layout so the numeric `InputGroupInput` is the only `FormControl`, and the sibling `Slider` is named via `aria-labelledby` pointing at the `FormLabel`. This removes the duplicate `${id}-form-item` id.
- Add a dev-time guard in `FormControl` that warns when the generated id lands on a missing or non-labelable element.

## Tests

- `packages/ui/src/components/form.test.tsx`
  - `InputGroup` control gets the generated id; `<fieldset>` does not.
  - Base UI `Slider` inside `FormControl` gets id and `aria-labelledby`.
  - Dual `Slider` + `InputGroup` layout keeps ids distinct.
  - Non-labelable wrapper triggers a dev warning.
- `apps/web/src/components/input/chip-input.test.tsx`
  - `ChipInput` inside `FormItem` wires the inner input to `FormLabel` and drops the generic `aria-label`.

## Verification

- `pnpm --filter @reactive-resume/ui test` — 35 files, 362 pass.
- `pnpm --filter web test` — 110 files, 595 pass.
- `pnpm --filter @reactive-resume/ui typecheck` — pass.
- `pnpm --filter web typecheck` — pass.
- `pnpm exec biome check --write` on changed files — clean.
- `python3 /home/ubuntu/Projects/open-source/scripts/preflight_ship.py --repo amruthpillai/reactive-resume --local reactive-resume --branch main --file packages/ui/src/components/form.tsx --must-contain 'id: `${id}-form-item`' --search '3369 in:title,body'` — clear.

## Visual changes

None (DOM wiring only).

## Follow-ups

The new dev guard surfaces a pre-existing dangling `for` in `RichTextField` / `URLInput` in the resume form. These are out of scope for this PR and are left for a separate pass.

Fixes #3369

## Summary by CodeRabbit

* **Accessibility**
  * Improved form control labeling and automatic ID associations across inputs, text areas, chip inputs, and sliders.
  * Improved propagation of validation and description attributes for assistive technologies.
  * Improved accessible labeling for the sidebar width slider.
  * Added development warnings when controls cannot be correctly associated with labels.

* **Bug Fixes**
  * Prevented redundant generic labels when a visible form label is available.
  * Ensured controls nested in input groups retain correct accessibility attributes.
  * Preserved custom slider data attributes when provided.

The PR repairs accessible label targeting when `FormControl` renders composite controls.
- Adds form-control context for forwarding generated IDs, labels, descriptions, and validation state to descendant controls.
- Routes InputGroup and ChipInput accessibility attributes to their inner inputs.
- Routes Slider labeling and validation attributes to its range inputs while preserving standalone root identity.
- Updates the sidebar-width field to avoid duplicate control IDs and adds regression coverage.
</details>

The PR appears safe to merge.

No blocking failure remains.
</details>

| Filename | Overview |
|----------|----------|
| packages/ui/src/components/form.tsx | Adds descendant control context, label IDs, and a development warning for dangling or non-labelable targets. |
| packages/ui/src/components/slider.tsx | Preserves standalone root IDs and transfers FormControl accessibility attributes to the actual range inputs, resolving the previous root-attribute regression. |
| packages/ui/src/components/input-group.tsx | Keeps generated form attributes off the fieldset and forwards them to the nested input or textarea. |
| apps/web/src/components/input/chip-input.tsx | Associates the internal text input with its visible FormLabel while retaining the generic fallback accessible name. |
| apps/web/src/routes/builder/$resumeId/-sidebar/right/sections/layout/index.tsx | Separates Slider labeling from the numeric FormControl to prevent duplicate IDs while naming both controls. |

</details>

<sub>Reviews (5): Last reviewed commit: ["fix(ui): remove internal label provider ..."](https://github.com/amruthpillai/reactive-resume/commit/f568b8feb9b4537b707020081469135f524e2c47) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=57528198)</sub>

### DIFF
--- a/packages/ui/src/components/form.tsx
+++ b/packages/ui/src/components/form.tsx
@@ -1,5 +1,4 @@
 import { useRender } from "@base-ui/react";
-import { LabelableProvider } from "@base-ui/react/internals/labelable-provider";
 import * as React from "react";
 import { Label } from "@reactive-resume/ui/components/label";
 import { cn } from "@reactive-resume/utils/style";
@@ -136,13 +135,7 @@ function FormControl({ render, ...props }: useRender.ComponentProps<"div">) {
 		},
 	});
 
-	return (
-		<FormControlContext.Provider value={contextValue}>
-			<LabelableProvider controlId={controlId} labelId={labelId}>
-				{element}
-			</LabelableProvider>
-		</FormControlContext.Provider>
-	);
+	return <FormControlContext.Provider value={contextValue}>{element}</FormControlContext.Provider>;
 }
 
 function FormDescription({ className, ...props }: React.ComponentProps<"p">) {
--- a/packages/ui/src/components/form.tsx
+++ b/packages/ui/src/components/form.tsx
@@ -1,4 +1,5 @@
 import { useRender } from "@base-ui/react";
+import { LabelableProvider } from "@base-ui/react/internals/labelable-provider";
 import * as React from "react";
 import { Label } from "@reactive-resume/ui/components/label";
 import { cn } from "@reactive-resume/utils/style";
@@ -12,6 +13,20 @@ const FormItemContext = React.createContext<FormItemContextValue>({ id: "", hasE
 
 function useFormItem() {
 	return React.use(FormItemContext);
+}
+
+type FormControlContextValue = {
+	id?: string;
+	labelId?: string;
+	hasError?: boolean;
+	"aria-describedby"?: string;
+	"aria-invalid"?: boolean | "true" | "false";
+};
+
+const FormControlContext = React.createContext<FormControlContextValue | null>(null);
+
+function useFormControl(): FormControlContextValue {
+	return React.use(FormControlContext) ?? {};
 }
 
 type FormItemProps = React.ComponentProps<"div"> & { hasError?: boolean };
@@ -32,6 +47,7 @@ function FormLabel({ className, ...props }: React.ComponentProps<typeof Label>) 
 
 	return (
 		<Label
+			id={`${id}-form-item-label`}
 			data-slot="form-label"
 			data-error={hasError}
 			className={cn("mb-0.5 data-[error=true]:text-destructive", className)}
@@ -41,23 +57,92 @@ function FormLabel({ className, ...props }: React.ComponentProps<typeof Label>) 
 	);
 }
 
+const LABELABLE_TAGS = new Set(["button", "input", "meter", "output", "progress", "select", "textarea"]);
+const LABELABLE_ROLES = new Set([
+	"button",
+	"combobox",
+	"gridcell",
+	"listbox",
+	"option",
+	"progressbar",
+	"radio",
+	"searchbox",
+	"slider",
+	"spinbutton",
+	"switch",
+	"tab",
+	"textbox",
+	"treeitem",
+]);
+
+function isLabelableElement(element: Element) {
+	if (LABELABLE_TAGS.has(element.tagName.toLowerCase())) return true;
+	if (element.getAttribute("contenteditable") === "true") return true;
+	const role = element.getAttribute("role");
+	if (role && LABELABLE_ROLES.has(role)) return true;
+	return false;
+}
+
+function useFormControlWarning(controlId: string) {
+	React.useEffect(() => {
+		if (process.env.NODE_ENV === "production") return;
+		if (typeof document === "undefined") return;
+
+		const element = document.getElementById(controlId);
+		if (!element) {
+			console.warn(
+				`FormControl: no element in the document has the generated id "${controlId}". The <FormLabel for="${controlId}"> target is dangling.`,
+			);
+			return;
+		}
+
+		if (!isLabelableElement(element)) {
+			console.warn(
+				`FormControl: the element that carries the generated id "${controlId}" is not a labelable element. A <label for="${controlId}"> will not name the control.`,
+			);
+		}
+	}, [controlId]);
+}
+
 function FormControl({ render, ...props }: useRender.ComponentProps<"div">) {
 	const { id, hasError } = useFormItem();
+	const controlId = `${id}-form-item`;
+	const labelId = `${id}-form-item-label`;
+	const describedBy = hasError ? `${id}-form-item-description ${id}-form-item-message` : `${id}-form-item-description`;
 
-	return useRender({
+	const contextValue = React.useMemo<FormControlContextValue>(
+		() => ({
+			id: controlId,
+			labelId,
+			hasError,
+			"aria-describedby": describedBy,
+			"aria-invalid": hasError,
+		}),
+		[controlId, describedBy, hasError, labelId],
+	);
+
+	useFormControlWarning(controlId);
+
+	const element = useRender({
 		defaultTagName: "div",
 		render,
 		state: { slot: "form-control" },
 		props: {
-			id: `${id}-form-item`,
+			id: controlId,
 			"data-slot": "form-control",
-			"aria-describedby": hasError
-				? `${id}-form-item-description ${id}-form-item-message`
-				: `${id}-form-item-description`,
+			"aria-describedby": describedBy,
 			"aria-invalid": hasError,
 			...props,
 		},
 	});
+
+	return (
+		<FormControlContext.Provider value={contextValue}>
+			<LabelableProvider controlId={controlId} labelId={labelId}>
+				{element}
+			</LabelableProvider>
+		</FormControlContext.Provider>
+	);
 }
 
 function FormDescription({ className, ...props }: React.ComponentProps<"p">) {
@@ -114,4 +199,4 @@ function FormMessage({ className, errors, ...props }: FormMessageProps) {
 	);
 }
 
-export { FormControl, FormDescription, FormItem, FormLabel, FormMessage };
+export { FormControl, FormControlContext, FormDescription, FormItem, FormLabel, FormMessage, useFormControl };
--- a/packages/ui/src/components/input-group.tsx
+++ b/packages/ui/src/components/input-group.tsx
@@ -9,14 +9,27 @@ import { cn } from "@reactive-resume/utils/style";
 
 function InputGroup({
 	className,
-	id: _id,
-	"aria-describedby": _ariaDescribedBy,
-	"aria-invalid": _ariaInvalid,
+	id: idProp,
+	"aria-describedby": ariaDescribedByProp,
+	"aria-invalid": ariaInvalidProp,
 	...props
 }: React.ComponentProps<"fieldset">) {
+	const formControl = useFormControl();
+	// Inside a FormControl composition the generated control props belong to
+	// the inner input (delivered via context), not the fieldset wrapper.
+	// Standalone usage has no FormControl, so explicit native fieldset props
+	// must survive untouched.
+	const insideFormControl = formControl.id !== undefined;
+	const fieldsetId = insideFormControl ? undefined : idProp;
+	const fieldsetDescribedBy = insideFormControl ? undefined : ariaDescribedByProp;
+	const fieldsetInvalid = insideFormControl ? undefined : ariaInvalidProp;
+
 	return (
 		<fieldset
 			data-slot="input-group"
+			id={fieldsetId}
+			aria-describedby={fieldsetDescribedBy}
+			aria-invalid={fieldsetInvalid}
 			className={cn(
 				"group/input-group relative flex h-9 w-full min-w-0 items-center rounded-lg border border-input outline-none transition-colors in-data-[slot=combobox-content]:focus-within:border-inherit in-data-[slot=combobox-content]:focus-within:ring-0 has-[>[data-align=block-end]]:h-auto has-[>[data-align=block-start]]:h-auto has-[>textarea]:h-auto has-[>[data-align=block-end]]:flex-col has-[>[data-align=block-start]]:flex-col has-[[data-slot=input-group-control]:focus-visible]:border-ring has-[[data-slot][aria-invalid=true]]:border-destructive has-disabled:bg-input/50 has-disabled:opacity-50 has-[[data-slot=input-group-control]:focus-visible]:ring-3 has-[[data-slot=input-group-control]:focus-visible]:ring-ring/50 has-[[data-slot][aria-invalid=true]]:ring-3 has-[[data-slot][aria-invalid=true]]:ring-destructive/20 has-[>[data-align=inline-start]]:[&>input]:ps-1.5 has-[>[data-align=inline-end]]:[&>input]:pe-1.5 has-[>[data-align=block-end]]:[&>input]:pt-3 has-[>[data-align=block-start]]:[&>input]:pb-3",
 				className,
--- a/packages/ui/src/components/input-group.tsx
+++ b/packages/ui/src/components/input-group.tsx
@@ -2,11 +2,18 @@ import type { VariantProps } from "class-variance-authority";
 import type * as React from "react";
 import { cva } from "class-variance-authority";
 import { Button } from "@reactive-resume/ui/components/button";
+import { useFormControl } from "@reactive-resume/ui/components/form";
 import { Input } from "@reactive-resume/ui/components/input";
 import { Textarea } from "@reactive-resume/ui/components/textarea";
 import { cn } from "@reactive-resume/utils/style";
 
-function InputGroup({ className, ...props }: React.ComponentProps<"fieldset">) {
+function InputGroup({
+	className,
+	id: _id,
+	"aria-describedby": _ariaDescribedBy,
+	"aria-invalid": _ariaInvalid,
+	...props
+}: React.ComponentProps<"fieldset">) {
 	return (
 		<fieldset
 			data-slot="input-group"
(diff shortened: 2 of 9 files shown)


packages/ui/src/components/slider.tsx
```diff
@@ -3,7 +3,19 @@ import { cn } from "@reactive-resume/utils/style";
...
-function Slider({ className, defaultValue, value, min = 0, max = 100, ...props }: SliderPrimitive.Root.Props) {
+function Slider({
+	className,
+	defaultValue,
+	value,
+	min = 0,
+	max = 100,
+	// Consumed by Base UI's LabelableProvider and applied to the real control
+	// input — re-applying it here would duplicate the id on the wrapper div.
+	id: _id,
```
amruthpillai: [P2] Preserve IDs supplied outside FormControl

This unconditionally removes the public `id` prop. The LabelableProvider receives an ID only when a FormControl supplies it, so `<Slider id="caller-slider" defaultValue={[30]} />` now renders no element with that ID; it exists on main. Preserve the explicit ID for standalone usage while avoiding duplicate generated IDs inside FormControl, and cover both compositions. This is a shared prop-contract regression; I did not find a current app call site supplying a standalone Slider ID.

amruthpillai (commented):
Verified the supplied accessibility tests and UI typecheck. Two standalone component compositions regress at this head: the same DOM probes pass on main and fail here. Details inline.

The remaining Basics → Website and Picture → Size cases from #3369 are covered by complementary PR #3424; that PR also fixes the shared WebsiteField. Please keep the issue scope tied to both changes.

packages/ui/src/components/input-group.tsx
```diff
@@ -2,11 +2,18 @@ import type { VariantProps } from "class-variance-authority";
...
 import { Input } from "@reactive-resume/ui/components/input";
 import { Textarea } from "@reactive-resume/ui/components/textarea";
 import { cn } from "@reactive-resume/utils/style";
 
-function InputGroup({ className, ...props }: React.ComponentProps<"fieldset">) {
+function InputGroup({
+	className,
+	id: _id,
+	"aria-describedby": _ariaDescribedBy,
+	"aria-invalid": _ariaInvalid,
```
amruthpillai: [P2] Preserve standalone fieldset identity and description

These props are discarded even without a FormControl ancestor. `<InputGroup id="caller-group" aria-describedby="hint"><InputGroupInput /></InputGroup>` loses both the fieldset ID and its description on this branch; both survive on main. Strip or transfer generated FormControl props only for that composition, preserving explicit native fieldset props for standalone callers. Please cover standalone and FormControl-wrapped cases; no current app use of a standalone InputGroup ID was found in this review.

packages/ui/src/components/slider.tsx
```diff
@@ -3,7 +3,17 @@ import { cn } from "@reactive-resume/utils/style";
...
-function Slider({ className, defaultValue, value, min = 0, max = 100, ...props }: SliderPrimitive.Root.Props) {
+function Slider({
+	className,
+	defaultValue,
+	value,
+	min = 0,
+	max = 100,
+	id: _id,
+	"aria-describedby": ariaDescribedBy,
+	"aria-invalid": _ariaInvalid,
```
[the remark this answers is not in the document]
amruthpillai: Fixed by 3cdd78a and retained in latest head. FormControl error state is applied to the real range input through inputRef; UI suite passes.

packages/ui/src/components/input-group.tsx
```diff
@@ -114,10 +121,24 @@ function InputGroupText({ className, ...props }: React.ComponentProps<"span">) {
...
 
-function InputGroupInput({ className, ...props }: React.ComponentProps<"input">) {
+function InputGroupInput({
+	className,
+	id: idProp,
+	"aria-describedby": ariaDescribedByProp,
+	"aria-invalid": ariaInvalidProp,
+	...props
+}: React.ComponentProps<"input">) {
+	const formControl = useFormControl();
```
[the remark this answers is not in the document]
amruthpillai: Verified current implementation. FormControl context centralizes generated id, label id, description, and invalid state; wrapper-specific consumers decide where native attributes belong to avoid duplicate ids. A generic merger would erase that DOM-specific distinction. Regression coverage now includes standalone and wrapped controls.

packages/ui/src/components/form.tsx
```diff
@@ -1,4 +1,5 @@
 import { useRender } from "@base-ui/react";
+import { LabelableProvider } from "@base-ui/react/internals/labelable-provider";
```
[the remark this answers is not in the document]
amruthpillai: Fixed in f568b8f. Removed the Base UI internal LabelableProvider dependency and bridged Slider id, label, description, and invalid state through public props/inputRef. UI 368 tests, web 596 tests, both typechecks, and Biome pass.

packages/ui/src/components/slider.tsx
amruthpillai: Fixed by de5bbf2. Standalone Slider preserves caller id, while single-thumb FormControl assigns generated id to the real input and range sliders keep unique input ids. Covered by regression tests.

packages/ui/src/components/input-group.tsx
amruthpillai: Fixed by de5bbf2. Standalone InputGroup preserves native identity/description props; FormControl composition transfers generated control props to the inner input. Both compositions are covered.

Issue #3152, opened 2026-06-08
[Bug] Enabling REDIS_URL and ENCRYPTION_SECRET pointing to adjacent fresh Redis server throws "relation "ai_providers" does not exist" errors, v5.1.4, self-hosted


BloodyIron:
### Is there an existing issue for this?

- [x] Yes, I have searched the existing issues and none of them match my problem.

### Product Variant

Self-Hosted

### Describe the bug you're experiencing

I am running Reactive Resume v5.1.4 in my Kubernetes cluster via YAML Deployment declaration, applied via ArgoCD, so no helm or konfig, etc. This is 100% self-hosted on-premises, I have 100% control over the entire environment it runs on and can have my servers touch grass if need be (or I can touch the servers, take your pick).

I'm also using postgresql as a container in the pod (loopback) using "postgres:latest" as the image version. Nothing more fancy than defining DB, User, and Password, and the PVC it uses for data pointing to /var/lib/postgresql .

I have been using Reactive Resume for a few months and have tried integrating it with my on-LAN Ollama AI endpoint. However I'm not getting traction.

When I enable the Environment Variables (and populate with appropriate values) REDIS_URL and ENCRYPTION_SECRET, I get oRPC Server DrizzleQueryErrors speaking to "relation "ai_providers" does not exist" as per the logs below. This gets outputted to the logs when I go to the "Integrations" page in my user profile to try to set up AI stuff.

### What template are you using?

None

### Anything else?

According to documentation this should "just work" and I have found zero resources online mentioning the ai_providers error produced.

I do not believe I have deviated from the documentation to any degree that SHOULD or COULD have caused this. So I am not currently seeing what I can do about this.

### LOGS (sorry):

```
[oRPC Server] DrizzleQueryError: Failed query: select "id", "user_id", "label", "provider", "model", "base_url", "encrypted_api_key", "api_key_salt", "api_key_hash", "api_key_preview", "test_status", "test_error", "last_tested_at", "last_used_at", "enabled", "created_at", "updated_at" from "ai_providers" where "ai_providers"."user_id" = $1 order by coalesce("ai_providers"."last_used_at", '1970-01-01T00:00:00.000Z'::timestamptz) desc, "ai_providers"."created_at" asc
params: 019c9b67-9b6a-71a8-8b9e-7735e5aa3fe2
    at file:///app/apps/web/.output/server/_libs/drizzle-orm.mjs:7224:11
[...]

PR #3513, opened 2026-09-12
fix(server): verify migrated schema at startup

santhiprakash:
Refs #3152

## Problem

On some self-hosted deployments the `ai_providers` table is missing at runtime even though startup migrations report success. Every AI providers route then throws a raw `DrizzleQueryError` wrapping Postgres `42P01` (`relation "ai_providers" does not exist`).

## Why migrations didn't prevent it

Investigated how this state can occur despite automatic startup migrations:

- Migrations run automatically at startup throughout the reporter's version range (Nitro plugin `apps/web/plugins/1.migrate.ts` in v5.1.x, `apps/server/src/startup/checks.ts` now), and a migration failure aborts boot. The reporter's logs show repeated `42P01` with no migration error — so startup succeeded.
- Drizzle's migrator selects pending migrations by folder name against `drizzle.__drizzle_migrations.name`; every release containing the `ai_providers` migration ships it, and migration folders were always committed in timestamp order, so neither current name-based nor older timestamp-based selection can skip it.
- The migration's DDL is complete and unmodified since it was introduced in PR #3062.

The consistent explanation is **ledger/schema divergence**: `drizzle.__drizzle_migrations` lives in the `drizzle` schema and records that a migration ran, but cannot detect objects dropped or lost outside the migrator — a partial backup restore, a manual `DROP TABLE`, or a recreated `public` schema while the `drizzle` schema survives. These happen on self-hosted Postgres instances outside the app's control.

## Fix

Verify the migrated schema at startup, right after `migrate()` succeeds:

- New `verifyMigratedSchema()` in `apps/server/src/startup/schema-check.ts` collects every table and column the Drizzle schema declares and checks them against the live catalog (`to_regclass` / `pg_attribute`) in one query.
- Missing objects throw an error naming each missing table or column; the existing `catch` in `runDatabaseMigrations` logs it and aborts boot — same failure surface as a real migration failure, matching the invariant that an unmigrated server must not serve traffic.
- Cost: one catalog query per boot.

## Verification

- `pnpm --filter server test` — 117 passed / 5 skipped, including 5 new tests in `schema-check.test.ts` (missing-table and missing-column failures, catalog param assertion, clean-catalog pass).
- `pnpm --filter @reactive-resume/api test` — 398 passed / 7 skipped (unchanged).
- `pnpm exec biome check` on the three changed files — clean (also via lefthook pre-commit).
- `pnpm --filter server typecheck` — fails only on pre-existing `packages/mcp/src/tools.ts` errors also present on upstream `main`; changed files produce no errors.

## Notes

- This replaces the earlier service-layer `42P01` → `PRECONDITION_FAILED` mapping: the `PRECONDITION_FAILED` UI banner hardcodes "requires `REDIS_URL` and `ENCRYPTION_SECRET` to be configured", which is misleading when those env vars are set but the table is missing. A startup failure naming the missing objects is more actionable.
- No schema or migration changes.

AI-assisted. I wrote and verified this change.


amruthpillai:
@santhiprakash Could you clarify how this state occurs through a supported deployment path? Migrations run automatically during startup, and a migration failure should prevent the server from starting.

If ai_providers is missing despite successful startup, we should investigate why that migration was skipped. This change improves the error message, but leaves the underlying migration problem unresolved. I’d prefer a reproduction and a fix at the migration/startup layer before adding service-specific handling.

santhiprakash:
@amruthpillai Good question — I dug into it and reworked the PR.

**Can this happen through a supported path?** I don't think the migrator itself can produce this state:

- Startup migrations have run automatically across the reporter's version range (Nitro plugin `apps/web/plugins/1.migrate.ts` in v5.1.x, `apps/server/src/startup/checks.ts` now), and a migration failure aborts boot. The reporter's logs show repeated `42P01` with no migration error — so startup succeeded.
- The migrator selects pending migrations by folder name against `drizzle.__drizzle_migrations.name`; the older timestamp-based selection can't skip it either, since migration folders were always committed in timestamp order. The `ai_providers` migration ships in every release in that range, unchanged since #3062.

The consistent explanation is **ledger/schema divergence**: `__drizzle_migrations` lives in the `drizzle` schema, so it survives a dropped/recreated `public` schema, a partial backup restore, or a manual `DROP TABLE`. The ledger says "applied" while the objects are gone — the migrator can't see drift it didn't cause, and the app serves traffic until the first query hits the missing table.

**Fix moved to the startup layer:** after `migrate()` succeeds, `verifyMigratedSchema()` compares every table/column the schema declares against the live catalog and aborts boot naming the missing objects — same failure surface as a real migration failure. One catalog query per boot.

I also dropped the service-layer `42P01` → `PRECONDITION_FAILED` mapping: the `PRECONDITION_FAILED` banner hardcodes "requires `REDIS_URL` and `ENCRYPTION_SECRET`", which would mislead this reporter — their env is configured; the table is gone. Failing startup with the missing objects named is more actionable.

Verified: `pnpm --filter server test` 117 pass (5 new), `pnpm --filter @reactive-resume/api test` 398 pass, biome clean on changed files. `pnpm --filter server typecheck` fails only on pre-existing `packages/mcp/src/tools.ts` errors that exist on `main` too.

santhiprakash:
Concrete reproduction of the divergent state, run end-to-end against real Postgres 16 with the repo's `migrations/` directory and drizzle-orm 1.0.0-rc.4 (the same `migrate()` the server calls at boot):

```
# 1. migrate() on a fresh database
MIGRATE: completed without error
to_regclass('ai_providers') = ai_providers
drizzle.__drizzle_migrations.name = '20260513181752_bent_human_cannonball'

# 2. lose the table outside the migrator
psql -c 'drop table "ai_providers" cascade'
#   (stand-in for any outside-migrator loss: public-schema-scoped restore,
#    manual drop, recreated public schema — the ledger in schema `drizzle` is untouched)

# 3. migrate() again — the ledger says applied, so nothing is re-created
MIGRATE: completed without error
to_regclass('ai_providers') = null        ← "successful" startup, table gone

# 4. the reporter's runtime error on a running server
select "id" from "ai_providers"
→ 42P01: relation "ai_providers" does not exist

# 5. the new startup check on the same database
verifyMigratedSchema() throws:
Database schema does not match the migration ledger: table "ai_providers"
missing even though all migrations are marked as applied. ...
```

So `migrate()` can neither detect nor repair this state — the ledger is its only source of truth — while the app would have booted and served traffic until the first AI-providers query.

One correction to my earlier note about the reporter's version: on v5.1.x the Nitro plugin invokes `migrate()` without awaiting it (`initNitroPlugins` calls `plugin(app)` synchronously and only catches sync throws), so the listener came up alongside the migration rather than gated on it — a contended or failing migration during a rolling update could surface `42P01` even without catalog drift. Current `main` does gate (`await runStartupChecks()` before `createApp`); the remaining gap is exactly the successful-migrate-but-drifted-catalog case this PR now fails fast on.

amruthpillai:
Verification follow-up: no actionable regression found, so no code changes proposed.

Compared head `719f4922a2fcefc6b8661d234e1db9f681cc85c4` with exact merge-base `fd3494ccac6ef4c47d01614a8a5dcdea08df85eb`, using matching lockfiles and installed dependency inventories.

- Server tests: head 117 passed / 5 skipped; base 112 passed / 5 skipped.
- Server typecheck (`--noErrorTruncation`): both fail with the same 526 diagnostics. Full outputs are byte-identical after normalizing worktree paths; zero new diagnostics, none in changed files.
- Node 24.21.0, pnpm 12.3.4. Installs used `--frozen-lockfile --ignore-scripts`.

This resolves the missing baseline comparison, not the existing typecheck failures. Five opt-in OAuth integration tests remained skipped; live PostgreSQL checks were not repeated in this verification.

amruthpillai (approved):
## Review — verified against a real database

Approving the approach; one trade-off needs a maintainer decision before merge (below). The diagnosis is sound: `drizzle.__drizzle_migrations` records that a migration *ran*, never that its objects still *exist*, so a partial restore / manual drop / recreated `public` schema is invisible and only surfaces later as `42P01`.

### Verification performed

Ran the real path end-to-end against a scratch `postgres:17`, not just the mocked `queryable`:

- fresh database + `migrate()` from this repo's `migrations/` → `verifyMigratedSchema(pool)` **resolves**. No false positive against the app's own migration output — this was the main regression risk and it is clear.
- `drop table ai_providers` → rejects with `table "ai_providers"`.
- dropping a single column of `resume_statistics` → rejects with `"resume_statistics"."<column>"`.
- bare catalog query on a clean schema returns 0 rows; `to_regclass('user')` handles the reserved-word table name correctly.
- `pnpm --filter server test` → 117 passed / 5 skipped.
- With the patch applied to a correctly installed checkout: `pnpm --filter server typecheck` reports **0** errors and `biome check` on the three files is clean.

### Decision needed

`apps/server/src/index.ts` awaits `runStartupChecks()`, and the entrypoint exits 1 on a throw. So this converts "AI provider routes return 42P01" into **"the server does not boot at all"**. That is the stated intent and the error message is genuinely actionable, but it is a severity escalation for self-hosters: a partially restored database now means a total outage instead of a degraded subset. Options, in my order of preference:

1. Gate it: `STRICT_SCHEMA_CHECK=true` for hard failure, otherwise a loud `console.error` and continue.
2. Keep the hard failure as written, accepting that a drifted database must be fixed before serving.
3. Downgrade to warning only — weakest option, since the original issue is precisely that the drift was silent.

Worth an explicit call in the PR body so it is a recorded decision either way.

### Nits

1. The throw is caught by `runDatabaseMigrations`'s `catch`, which logs `"Database migrations failed"`. A post-migration verification failure logged as a migration failure is misleading — either log a distinct message there or move the call outside that `try`.
2. `to_regclass(e.table_name)` resolves through `search_path`. I confirmed a table outside the path is reported as missing. All current tables are unqualified in `public`, so this works today, but a self-hoster that sets `search_path` in their connection options gets a false-positive boot failure. Qualifying the name (or comparing against `to_regclass('public.' || name)`) removes the foot-gun.
3. `collectExpectedColumns()` covers tables and columns only — no indexes, constraints, or enums. That is a fine scope; the module comment should say so explicitly so nobody assumes it is a full schema diff.
4. The query relies on `$1`/`$2` being index-aligned for the `unnest`. A one-line comment naming that invariant would help the next reader.
5. `schema-check.test.ts`: the "passes the expected table and column lists" case asserts `tables).toHaveLength(columns.length)`, which is close to asserting the implementation. The behavioural cases (missing table, missing column, clean pass) carry the weight, so this is optional.

santhiprakash:
Thanks for the thorough review — all addressed in `b8fee491` (rebased onto current `main` at `4fde62df6`, no overlapping upstream changes).

**Decision:** went with option 1 — the check is gated behind `STRICT_SCHEMA_CHECK`. Drift now logs `Database schema verification failed` with the missing objects named, and boot continues; `STRICT_SCHEMA_CHECK=true` makes it fatal. Wired through `packages/env/src/server.ts` (`z.stringbool().default(false)`), `turbo.json` `globalEnv`, and documented in `.env.example`. Recorded in the PR body under "Boot severity (decision)".

**Nits:**
1. `verifyMigratedSchema` now runs in its own `try` after the migration block, so drift logs as `Database schema verification failed`, not a migration failure. `pool.end()` still runs in the shared `finally`.
2. Both `to_regclass` calls now resolve `'public.' || e.table_name` — no `search_path` dependence.
3. The module comment now states the scope is tables and columns only (no indexes/constraints/enums).
4. Added a comment noting `$1`/`$2` are index-aligned for the `unnest`.
5. Replaced the `toHaveLength` assertion with an index-alignment check (`ai_providers.user_id` pairing) — same invariant as nit 4.

Verified on the rebased head:
- `pnpm --filter server test` — 118 passed / 5 skipped on Node 24 (one unrelated `openapi/generator.test.ts` cold-generation case can exceed its inline timeout under parallel load on this machine; passes standalone in 3.8s, no overlap with this diff)
- `pnpm --filter server typecheck` — 0 errors
- `pnpm exec biome check` on changed files — clean

PR #3521, opened 2026-09-16
fix(resume): detect section headings set in a side column

### LINKED ISSUE #3519: bronzor template does not pass deep check
### Existing issue

- [x] I searched the existing issues and could not find a matching report.

### Product variant

Cloud

### Reactive Resume version

5.3.0

### Area

Templates, preview & export

### Environment

Windows 11, Vivaldi

### Summary

when having the bronzor template selected, running ats checks returns 100% but the deep check returns 60%. however, any other template returns 100% on deep check for the exact same content.
i have also tried to remove all icons or export the cv in json and reimport it but its the same.

### Steps to reproduce

1. have the bronzor template selected
2. write the resume (making sure all required sections are filled with title, date, description, etc.)
3. run deep check

### Expected behavior

what is expected is to have deep check return 100%, just like it does on any other template.

```
Sections
100
Whether the resume is segmented the way software expects. 9 of 9 applicable checks passed.

Nothing to fix here.
```

### Actual behavior

the actual behaviour is having deep check return 60% on bronzor template because it cannot read the existing work experience, skills and education sections.

```
Sections
5
4 to fix
Whether the resume is segmented the way software expects. 4 of 8 applicable checks passed.

No work-experience section was found.
Add a section headed "Experience" or "Work Experience". It is the field most systems rank on.

No education section was found.
Add a section headed "Education". Many systems have a required education field.

No conventional section headings were found.
Use plain headings such as "Experience", "Education" and "Skills".

No skills section was found.
List the tools and technologies you use. Recruiters search skills more than any other field.
```

### Template

Bronzor

### Logs and screenshots

Image

santhiprakash:
## Problem

With the **bronzor** template selected, the resume health check's deep tier reports **~60** for the same resume content that scores 100 on every other template (and 100 on the standard ATS check). Fixes #3519.

## Triage / Root cause

The deep check renders the resume to a real PDF and analyses the extracted text geometry (`packages/resume/src/ats-pdf`). Bronzor sets each section title in a narrow column beside the section's content, so the title and the first content line share a baseline. `buildExtractedDocument` clusters spans by baseline, which merges the pair into a single line — e.g. `Profiles GitHub`, `Summary Passionate game developer with 5+ years…`, `Skills Unity Engine`, `Education University of Washington Bachelor of Science • 3.6 GPA`. The merged line no longer looks like a heading, so `detectHeadings` finds nothing, `sectionTypes` comes back empty, and the section checks fail together (`NO_RECOGNIZED_HEADINGS` / `FEW_SECTION_HEADINGS` / `NO_EXPERIENCE_SECTION` / `NO_EDUCATION_SECTION` / `NO_SKILLS_SECTION` / `NO_SUMMARY_SECTION`) — roughly the 40-point drop reported.

## Fix

- `extract.ts`: new `columnSegments(line)` view that splits a clustered line at column-sized gaps between its spans (> 20pt and > 2.5× font size). Lines without such gaps are returned unchanged, so single-column layouts are unaffected.
- `analyze/sections.ts`: `detectHeadings` evaluates each column-separated segment instead of the whole merged line, so a title set in its own column is recognised again.
- No other consumer changes: nothing slices section content by heading `lineIndex`, and every other analysis keeps using the merged lines.

## Verification

Sabotage-checked: both new tests fail on current `main` and pass with the fix.

- `pnpm --filter @reactive-resume/pdf exec vitest run src/ats-extraction.integration.test.tsx` — new test renders a real bronzor PDF via `@react-pdf/renderer` + pdf.js; on `main` it fails (`FEW_SECTION_HEADINGS` + 10 other check failures), with the fix the export scores 100. 13/13 pass.
- `pnpm --filter @reactive-resume/resume exec vitest run src/ats-pdf` — new unit test builds a bronzor-shaped extraction (title at x=33, content at x=201 sharing a baseline); on `main` `NO_RECOGNIZED_HEADINGS` fails. 117/117 pass.
- `pnpm --filter @reactive-resume/resume test` — 1381/1381; `pnpm --filter @reactive-resume/pdf test` — 1083/1083.
- `pnpm --filter @reactive-resume/resume typecheck` and `--filter @reactive-resume/pdf typecheck` (tsgo) clean; `pnpm exec biome check` on changed files clean.

## Notes / Risks

- The split threshold (20pt / 2.5× font size) is deliberately wider than the existing space-insertion gap (0.25× font size) so ordinary wide kerning can't fragment a line; only genuine column gaps split.
- Right-aligned runs sharing a baseline with left-aligned text (e.g. `Company … Jan 2020`) also split, which is correct — each segment is evaluated on its own text.
- Other templates verified unchanged: onyx and kakuna still score 100, gengar 97 (its pre-existing gutter warning is untouched).

## Assistance

AI-assisted. I wrote and verified this change.


amruthpillai (approved):
## Review — verified against the branch

Approving. Root cause is correct and the fix is aimed at it properly: `clusterLines` groups spans by baseline only, so bronzor's side-column title merges with the first content line of the section (`Profiles GitHub`), `detectHeadings` then finds no heading, and the six section checks fail together.

### Verification performed

- **Sabotage check** (revert only the two source files, keep the tests): `rules.test.ts` fails on `main` with `NO_RECOGNIZED_HEADINGS`; the bronzor integration test fails on `main` with `FEW_SECTION_HEADINGS` plus 10 findings. Both new tests are real regression guards, not vacuous.
- **Suites on the branch**: `@reactive-resume/resume` 1381/1381, `@reactive-resume/pdf` 1038/1038.
- **15-template sweep, `main` vs branch** — real PDFs rendered through `@react-pdf/renderer` + pdf.js, same analyzer, `sampleResumeData`:

```
azurill    98 → 100     chikorita  95 → 97     glalie   95 → 97
bronzor    60 → 100     ditgar     95 → 97     leafish  98 → 100
ditto 97→97   gengar 97→97   kakuna 100→100   lapras 100→100   meowth 100→100
onyx 100→100  pikachu 97→97  rhyhorn 100→100  scizor 100→100
```

Every delta is a removed false section finding; no template gains a finding, and bronzor lands on exactly the 100 the issue asks for. That is the strongest possible evidence for a change of this shape.

### Code

- Layering is right: `columnSegments` is a layer-2 view on `ExtractedDocument` consumed by layer-3 `analyze`, and it returns the original line for `spans.length < 2` and for a single group, so single-column layouts are untouched.
- The split threshold (≥ 20pt **and** ≥ 2.5× font size) sits far above the 0.25× font size space-insertion gap, so kerning cannot fragment a line. Defining the constants next to `SPACE_GAP_RATIO` with the contrast stated is the right place.
- `headings[].lineIndex` keeps its meaning; nothing consumes heading `lineIndex` today (`dates.ts` computes its own), so two headings sharing a merged line's index is harmless.

### Non-blocking

1. Only heading detection reads the split view. The merged `line.text` (`Profiles GitHub`) still feeds text-quality, contact, and date analysis. That is acceptable — those checks are content-oriented and the sweep shows no fallout — but the *root* fix is to stop merging spans across a column-sized gap in `clusterLines` itself. That would touch every consumer and change `line.text` for all templates, so the surgical approach here is the right call for now. Worth a follow-up issue rather than a change in this PR.
2. The bronzor integration test asserts absent codes rather than `score === 100` (the report does score 100). An explicit score assertion would be a slightly stronger guard.
3. If a future consumer ever slices section content by heading `lineIndex`, document that a column-split heading shares its index with the merged line.

